PR #60, opened 2022-10-28
feat(base-cluster): lower base resource usage

cwrau:


marvinWolff (approved):
Let's hope this will be enough so that it can be deployed.

marvinWolff (approved):

PR #100, opened 2022-12-06
feat(base-cluster): add rbac configuration options

### LINKED ISSUE #99: [base-cluster] user management
We should be able to manage various user accounts via the helm chart

we need the following;

- [x] list of user accounts
- [x] (Cluster-)Roles?
- [x] (Cluster-)Rolebindings?
- [ ] Namespaces?

~We probably can't automate the creation of the tokens because of the "recent" changes in k8s v1.23~

We can just create the secret manually 🤷

~Therefore~ Still, we should print a command that generates the kubeconfig, including the creation of the token

cwrau:
allows creating roles and accounts and binding the roles to the accounts either per namespace or clusterwide

closes #99


cwrau:
@marvinWolff during development it occurred to me, that maybe we should also provide a way to create namespaces?

Otherwise the `RoleBinding`s for the various namespaces can't be created until the corresponding namespace is somehow created 🤷

marvinWolff (dismissed):

marvinWolff:
Yes, that's a good idea.

charts/base-cluster/templates/monitoring/kube-prometheus-stack/_config.yaml
```diff
@@ -173,6 +173,7 @@ prometheusOperator:
 defaultRules:
   disabled:
     etcdHighNumberOfFailedGRPCRequests: true
+    etcdDatabaseHighFragmentationRatio: true
```
marvinWolff, line 176: So we disable the etcd alerts and not fix them?

cwrau: Ups, this slid in 🤭

marvinWolff (approved):

PR #124, opened 2023-01-10
feat(base-cluster): enable brotli and gzip

marvinWolff:
As this has no downside we enable brotli and gzip compression for the nginx.


charts/base-cluster/templates/ingress/nginx.yaml
```diff
@@ -29,6 +29,8 @@ spec:
           additionalLabels: {{- .Values.monitoring.labels | toYaml | nindent 12 }}
       config:
         use-proxy-protocol: "true"
+		use-gzip: "true"
+		enable-brotli: "true"
```
cwrau, line 33: Use spaces instead of tabs

marvinWolff: Done.

Issue #61, opened 2022-10-28
[base-cluster] too high default resource requests


cwrau:

cwrau:
As we currently don't have *any* such requests / limits we should prioritize this

cwrau:
#151 wasn't enough, things like descheduler, cert-manager, kyverno (although deprecated and to be deactivated), metrics-server, ... request way too much

PR #151, opened 2023-01-25
feat(base-cluster): lower base resource requests

cwrau:
Relates to #61


marvinWolff:
Have you tested these values?

Looks good, but i can't say if those requests are realistc.

marvinWolff (approved):

PR #214, opened 2023-02-21
feat(ci/diff): push to dpaste and insert link

### LINKED ISSUE #168: [ci] upload PR diffs to pastebin
GitHub has a limit on comments lengths, and our Diffs easily break them 😅

cwrau:
Closes #168

### DIFF
--- a/.github/scripts/create-values-diff.sh
+++ b/.github/scripts/create-values-diff.sh
@@ -81,7 +81,7 @@ function generateComment() {
 
 
     diffs+=(
-      [$values]="$(diff -ur "$originalResourcesDir" "$newResourcesDir")"
+      [$values]="$(diff -ur "$originalResourcesDir" "$newResourcesDir" | curl -s -F syntax=diff -F "content=<-" https://dpaste.com/api/v2/)"
     )
     break
   done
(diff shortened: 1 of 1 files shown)


.github/scripts/create-values-diff.sh
```diff
@@ -81,7 +81,7 @@ function generateComment() {
 
 
     diffs+=(
-      [$values]="$(diff -ur "$originalResourcesDir" "$newResourcesDir")"
+      [$values]="$(diff -ur "$originalResourcesDir" "$newResourcesDir" | curl -s -F syntax=diff -F "content=<-" https://dpaste.com/api/v2/)"
```
marvinWolff, line 84: I would say i'm not a fan of pushing our diffs somewhere in the cloud, but we're doing it already :sweat_smile: 

Anyway, how long will the diff be available? When i visited the site the default was one week.

cwrau: Yeah, the default is a week, though we could raise that up to a whole year

I would say a PR ideally shouldn't lie around for more than a week, tho?

marvinWolff: Hm, not sure.  

This is the diff of the PR, right?

marvinWolff: But actually, what would we want to do with them.  

When something breaks, we will most likely notice it shortly after, and if not there is still `git blame` to see who added the break when.

Let's keep it a week.

marvinWolff (approved):

Issue #205, opened 2023-02-16
[deps/renovate] try detecting subchart versions


cwrau:
Maybe with https://docs.renovatebot.com/configuration-options/#regexmanagers

PR #216, opened 2023-02-21
feat(t8s-cluster): move version so renovate can detect this

cwrau:
Relates to #205


cwrau:
This won't work in the t8s-cluster, closing for now

PR #302, opened 2023-03-21
fix(t8s-cluster)!: wrong type for customerID, should be int

cwrau:


marvinWolff (approved):

cwrau:
> Actually it might be useful to have this as a string. One never knows what will be used as customerID. There might as well be a UUID or something like that. Since it's only an internal reference, the data type doesn't really matter for program logic, but limiting it to integers might be a problem in the future, whereas you totally could represent normal integers with a string.

As our operator has this field as type `int` we should change either the operator to `string` or the Helm Chart to `int`.

I was thinking the operator would be correct

marvinWolff (approved):

PR #322, opened 2023-03-27
feat(t8s-cluster): prepare for flux drift-detection

cwrau:


cwrau:
> I do get, what drift detection is. I also do approve choosing a bigger interval for helm-releases. Especially since 1m is pretty fast.
> 
> What I don't get is, what's the connection between longer intervals and drift detection. To me, these are two independent things.

Currently the interval doesn't do anything, therefore even `1s` wouldn't be a problem. But in the, hopefully near, future flux will _really_ reconcile the HelmReleases at the specified interval.

This of course causes semi-high load. And as we don't really expect there to be many random deletion / changes of the created resources it's good to have a semi-long interval as not to overload flux.

marvinWolff:
You mean because the drift detection would do the dry-run on every reconcile?

Changes look good to me :+1:

Issue #418, opened 2023-05-10
[t8s-cluster] configure kubelet to pull images in parallel


cwrau:
Only after 1.27 adoption, see https://kubernetes.io/docs/concepts/containers/images/#maximum-parallel-image-pulls

PR #441, opened 2023-06-07
feat(t8s-cluster): Allow for parallel image pulling with k8s v1.27

ghost:
This allows to configure a max number of images pulled in parallel and turns off serializeImagePulls in that case. This only works with k8s v1.27 and up. The default is keept to 1 for now to test the functionality on specific clusters over a periode of time. It will probabbly be raised later on.

References:
- #418

### DIFF
--- a/charts/t8s-cluster/values.yaml
+++ b/charts/t8s-cluster/values.yaml
@@ -17,7 +17,7 @@ global:
       tag: 1.27.2
   injectedCertificateAuthorities: ""
   kubeletExtraConfig:
-    # Only use this with k8s v1.27 and up
+    # This is only used when using 1.27 or later
     maxParallelImagePulls: 4
 
 metadata:


charts/t8s-cluster/values.yaml
```diff
@@ -16,6 +16,9 @@ global:
       repository: bitnami/kubectl
       tag: 1.27.2
   injectedCertificateAuthorities: ""
+  kubeletExtraConfig:
+    # Only use this with k8s v1.27 and up
+    maxParallelImagePulls: 1
```
cwrau, line 21: Maybe we could set the default to be higher?

I can barely think of cases where you *wouldn't* want parallel image pulls

We could maybe do 4, 2 parallel pulls per core or something like that and then maybe just a flag to turn off the parallelization?

charts/t8s-cluster/values.yaml
```diff
@@ -16,6 +16,9 @@ global:
       repository: bitnami/kubectl
       tag: 1.27.2
   injectedCertificateAuthorities: ""
+  kubeletExtraConfig:
+    # Only use this with k8s v1.27 and up
```
cwrau, line 20:
```suggestion
    # This is only used when using 1.27 or later
```

cwrau (approved):

PR #475, opened 2023-07-07
feat(chirpstack): Add opinionated chirpstack chart

ghost:
This adds an helmchart to install chirpstack.

This is still a WIP, but feedback is already welcome.

Some of the things that still need to be done:
- Add a README.md explaining what this does and how to use it
- Add a License
- Change the default DB to a StatefulSet
- Include with existing CI jobs


cwrau:
> Add a License

By default everything is [MIT](https://github.com/teutonet/teutonet-helm-charts/blob/main/LICENSE)

> Include with existing CI jobs

I don't know which jobs you mean specifically, but all charts are built automatically

charts/chirpstack/templates/chirpstack/ingress.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if .Values.ingress.enabled -}}
+{{- if false }}
```
marvinWolff, line 2: What exactly is checked here? :sweat_smile:

charts/chirpstack/templates/postgres/_helpers.tpl
```diff
@@ -0,0 +1,16 @@
+{{- define "chirpstack.postgres.env" -}}
+- name: POSTGRES_PASSWORD
+  value: root
```
marvinWolff, line 3: Ehm, is this intended?

charts/chirpstack/templates/chirpstack/ingress.yaml
cwrau: This is for the IDE, in our case IntelliJ, so that it knows what to autocomplete

charts/chirpstack/templates/chirpstack-gateway-bridge/configmap.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: ConfigMap
+metadata: {{- include "chirpstack.commonMetadata" (dict "component" "gateway-bridge" "context" .) | nindent 2 }}
+data:
+  chirpstack-gateway-bridge.toml: |
+    # See https://www.chirpstack.io/gateway-bridge/install/config/ for a full
+    # configuration example and documentation.
+
+    [integration.mqtt.auth.generic]
+    servers=["tcp://{{ include "chirpstack.fullname" (dict "context" . "component" "mosquitto") }}:1883"]
+    username=""
```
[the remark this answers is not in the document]
marvinWolff: Probably not, i just saw the hardcoded empty values and thought maybe this should be configurable. But it should be fine as long as it's not exposed and the internal mqtt is used.

charts/chirpstack/templates/chirpstack-gateway-bridge/service.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
marvinWolff: Alright :)

charts/chirpstack/templates/postgres/_helpers.tpl
marvinWolff: Alright. Yes, you're right, it probably is confusing if we require two passwords.  
There is still the possibility to generate the password, but i guess using the postgres operator would make that obsolete.

marvinWolff:
LGTM :+1:

marvinWolff (approved):

charts/chirpstack/README.md.gotmpl
```diff
@@ -0,0 +1,45 @@
...
+For this to work, you will have to provide some configuration through your
+`values.yaml` (or set them through CLI).
+
+```yaml
+chirpstack:
+  # This secret is used for generating login and API tokens, make sure this
+  # is never exposed. Changing this secret will invalidate all login and API
+  # tokens. The following command can be used to generate a random secret:
+  #   openssl rand -base64 32
+  apiSecret: CHANGEME
```
cwrau, line 33: I would say we should by default generate a value for this so the user doesn't have to think about this and also doesn't have to add secrets to their git repo when using gitops

charts/chirpstack/templates/postgres/statefulset.yaml
```diff
@@ -0,0 +1,58 @@
+{{- if not .Values.postgres.existingPsqlUrl -}}
```
cwrau, line 1: I don't think we should deploy/manage our own postgres, we should either use an operator for this or force the user to give us an existing database connection

charts/chirpstack/templates/mosquitto/deployment.yaml
```diff
@@ -0,0 +1,45 @@
...
+      app.kubernetes.io/component: mosquitto
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels: {{- include "common.labels.standard" . | nindent 8 }}
+        app.kubernetes.io/component: mosquitto
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/mosquitto/configmap.yaml") . | sha256sum }}
+    spec:
```
cwrau, line 17: You're missing the podSecurityContext

charts/chirpstack/templates/postgres/configmap.yaml
```diff
@@ -0,0 +1,22 @@
...
+metadata: {{- include "chirpstack.commonMetadata" (dict "context" . "component" "postgres") | nindent 2 }}
+data:
+  001-init-chirpstack.sh: |
+    #!/bin/bash
+    set -e
+
+    psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" <<-EOSQL
+        create role $CS_POSTGRESQL_USER with login password '$CS_POSTGRESQL_PASSWORD';
+        create database $CS_POSTGRESQL_DB with owner $CS_POSTGRESQL_USER;
+    EOSQL
```
cwrau, line 13: Is this an extra user/db? Otherwise, aren't there any flags to auto-create this?

But this is less important than https://github.com/teutonet/teutonet-helm-charts/pull/475#discussion_r1275830328

charts/chirpstack/templates/postgres/statefulset.yaml
```diff
@@ -0,0 +1,58 @@
...
+    matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
+      app.kubernetes.io/component: postgres
+  serviceName: {{ include "chirpstack.fullname" (dict "context" . "component" "postgres") }}
+  template:
+    metadata:
+      labels: {{- include "common.labels.standard" . | nindent 8 }}
+        app.kubernetes.io/component: postgres
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/postgres/configmap.yaml") . | sha256sum }}
+    spec:
```
cwrau, line 17: Missing podSecurityContext

charts/chirpstack/templates/postgres/statefulset.yaml
```diff
@@ -0,0 +1,58 @@
...
+      {{- include "chirpstack.securityOptions" (dict) | nindent 6 }}
+      containers:
+        - env: {{- include "chirpstack.postgres.env" . | nindent 12 }}
+          image: {{ include "common.images.image" (dict "imageRoot" .Values.postgres.image "global" .Values.global) }}
```
cwrau, line 21: Missing containerSecurityContext

charts/chirpstack/templates/redis/deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+  selector:
+    matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
+      app.kubernetes.io/component: redis
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels: {{- include "common.labels.standard" . | nindent 8 }}
+        app.kubernetes.io/component: redis
+    spec:
```
cwrau, line 15: Missing podSecurityContext

charts/chirpstack/templates/redis/deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+      {{- include "chirpstack.securityOptions" (dict) | nindent 6 }}
+      containers:
+        - image: {{ include "common.images.image" (dict "imageRoot" .Values.redis.image "global" .Values.global) }}
+          resources: {{- .Values.redis.resources | toYaml | nindent 12 }}
```
cwrau, line 19: Missing containerSecurityContext

charts/chirpstack/templates/redis/persistentVolumeClaim.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
```
cwrau, line 1: Does the redis even need to be persistent? 🤔

Often it's just a cache

cwrau (changes requested):

PR #488, opened 2023-08-07
feat(base-cluster/nginx): enable underscores in headers

marvinWolff:
In preparation of the migration of customer clusters this change is needed.

### DIFF
--- a/charts/base-cluster/templates/ingress/nginx.yaml
+++ b/charts/base-cluster/templates/ingress/nginx.yaml
@@ -26,6 +26,7 @@ spec:
         use-proxy-protocol: "true"
         use-gzip: "true"
         enable-brotli: "true"
+        enable-underscores-in-headers: true
       service:
         annotations:
           loadbalancer.openstack.org/proxy-protocol: "true"


charts/base-cluster/templates/ingress/nginx.yaml
```diff
@@ -26,6 +26,7 @@ spec:
         use-proxy-protocol: "true"
         use-gzip: "true"
         enable-brotli: "true"
+        enable-underscores-in-headers: true
```
cwrau, line 29: Did you make sure that this works as a boolean? The other flags are strings

I'd rather have a real boolean, but you should make sure it works and if it does change the other flags to booleans as well.

marvinWolff: It is a bit confusing, i would say real bool should be right.  
If you look into the configmap values: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/
you can see, that for example brotli is also declared `bool`, but the value has double quotes. In contrast the underscores option is declared the same, but the value has no double quotes.  

So i guess it should be right?

marvinWolff:
> What does this actually do? I don't find the explanation in the documentation helpful to be honest.
> 
> https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#enable-underscores-in-headers
> 
> > Enables underscores in header names. default: is disabled
> 
> Or let me rephrase this: Why is this even an option and not just working? And why do we need it?
> 
> From a technical standpoint, this looks fine.

This is needed for the migration of one customer. It is valid, but not enabled by default in nginx, for whatever reason. I thought it would not hurt to enable this by default everywhere.

charts/base-cluster/templates/ingress/nginx.yaml
marvinWolff: Just tested it, works without problems like this.

PR #491, opened 2023-08-16
feat(base-cluster/velero): increase velero resources

marvinWolff:


marvinWolff:
> The `nodeAgent` limits are not needed?

They are, i've just added them

PR #495, opened 2023-08-17
fix(base-cluster/backups): Allow for configuring s3 buckets not in minio

ghost:
Adding "aws" as a new valid type of provider for backupStorageLocations. There's a lot of similarities with "minio", and to be honest I'd like to replace the "minio" variant entirely. But for backwards compatibility reasons, this was added as a new provider.

### DIFF
--- a/charts/base-cluster/templates/backup/velero.yaml
+++ b/charts/base-cluster/templates/backup/velero.yaml
@@ -81,7 +81,7 @@ spec:
           {{- $providerSpec := get $spec.provider $providerName }}
           config:
             region: {{ ($providerSpec.region | default "Region1") | quote }}
-            s3ForcePathStyle: {{ ($providerSpec.forcePathStyle | default "true") | quote }}
+            s3ForcePathStyle: {{ $providerSpec.forcePathStyle | default true }}
             s3Url: {{ $providerSpec.url | quote }}
           {{- end }}
       {{- end }}


charts/base-cluster/templates/backup/velero.yaml
```diff
@@ -77,6 +77,17 @@ spec:
...
+          {{- if eq $providerName "aws" }}
+          {{- $providerSpec := get $spec.provider $providerName }}
+          config:
+            {{- if $providerSpec.region }}
+            region: {{ $providerSpec.region | quote }}
+            {{- else }}
+            region: Region1
+            {{- end }}
+            s3ForcePathStyle: {{ ($providerSpec.forcePathStyle | default "true") | quote }}
+            s3Url: {{ $providerSpec.url | quote }}
```
cwrau, line 89: Those options are explicitly *not* for aws but for something like minio as aws does not have different URLs as is also referenced in the [velero config](https://velero.io/docs/v1.1.0/api-types/backupstoragelocation/#aws)

You could, and should, rather just adjust the minio provider to have these options, including `region` I guess

cwrau (changes requested):

charts/base-cluster/templates/backup/velero.yaml
```diff
@@ -77,6 +77,13 @@ spec:
             name: {{ get $existingSecret "name" | required (printf "You need to provide the existingSecret.name for %s" $name) }}
             key: {{ get $existingSecret "key" | default (printf "%s-%s" $providerName $name) }}
           {{- end }}
+          {{- if eq $providerName "minio" }}
+          {{- $providerSpec := get $spec.provider $providerName }}
+          config:
+            region: {{ ($providerSpec.region | default "Region1") | quote }}
+            s3ForcePathStyle: {{ ($providerSpec.forcePathStyle | default "true") | quote }}
```
cwrau, line 84: This is a boolean, are you sure this has to be quoted? Does it even work quoted? 🤔

charts/base-cluster/templates/backup/velero.yaml
cwrau: > I explicitly didn't touch the `minio` configuration to keep backwards compatibility. But I can, off course, add this to `minio` as well. Although I got to admit, that I find the name `minio` a bit misleading, as this configuration is not specific to minio but to most S3 backends (even AWS).

Yeah, we could also rename `minio` to something else if you got something, maybe `genericS3`? 🤔

But the `s3url` is not valid for real aws which is why I want them to be seperate

cwrau (changes requested):

cwrau:
> > Yeah, we could also rename minio to something else if you got something, maybe genericS3? 🤔
> > But the s3url is not valid for real aws which is why I want them to be seperate
> 
> I'd rather have it the same way, as it is used in velero. Which uses `aws` and makes the `url` optional, as it's only useful for minio and stuff. It even states, that you still could configure this, if you use the real AWS. It's just not useful, as you would have to create the full URL yourself, when the automatic can do it better.
> 
> -- | -- | -- | --
> 
> s3Url	string	Required field for non-AWS-hosted storage	Example: http://minio:9000 You can specify the AWS S3 URL here for explicitness, but Velero can already generate it from region, and bucket. This field is primarily for local storage services like Minio.
> See [velero.io/docs/v1.1.0/api-types/backupstoragelocation/#aws](https://velero.io/docs/v1.1.0/api-types/backupstoragelocation/#aws)
> 
> However, changing the name of this now, would be a breaking change. This is why I added an additional variant in the first place.

Mh, I understand what you mean, but we're providing an abstraction over the software we're deploying for the user in which case I'd probably want to group the various (required) configuration options for better input validation.

The `s3url` shouldn't be used for aws, whereas it's required, together with `s3ForcePathStyle` (`=true` in most cases, maybe even all? 🤷), for minio and other S3 providers.

And some S3 providers ignore the region completely while it's kinda required for aws.

If we give the user more chances to create a fundamentally invalid configuration then it's more likely to happen, which we can already observe internally, which was a major argument for this chart.

If we can filter out various invalid configuration for the user I'd say that's good, no?

cwrau (approved):

PR #545, opened 2023-09-01
feat(base-cluster/monitoring): monitor all namespaces by default

cwrau:
also add flag to only monitor rolled out namespaces

### DIFF
--- a/charts/base-cluster/templates/monitoring/kube-prometheus-stack/_prometheus_config.yaml
+++ b/charts/base-cluster/templates/monitoring/kube-prometheus-stack/_prometheus_config.yaml
@@ -43,11 +43,17 @@ prometheusSpec:
   probeSelector:
     matchLabels: {{- .Values.monitoring.labels | toYaml | nindent 6 }}
   additionalAlertRelabelConfigs:
+    {{- if not .Values.monitoring.monitorAllNamespaces }}
+      {{- $namespaces := list .Release.Namespace "kube-node-lease" "kube-public" "kube-system" -}}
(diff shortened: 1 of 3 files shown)


charts/base-cluster/templates/monitoring/kube-prometheus-stack/_prometheus_config.yaml
```diff
@@ -43,11 +43,17 @@ prometheusSpec:
   probeSelector:
     matchLabels: {{- .Values.monitoring.labels | toYaml | nindent 6 }}
   additionalAlertRelabelConfigs:
+    {{- if .Values.monitoring.monitorAllNamespaces }}
+      {{- $namespaces := list .Release.Namespace "kube-node-lease" "kube-public" "kube-system" -}}
```
marvinWolff, line 47: So you can just monitor all or none?  
Or did i miss something?

marvinWolff: I think a seperation for the three namespaces manually added to the list and all other created ones would be good. Or maybe the possibility to simply pass a list,  overwriting the flag.

marvinWolff (approved):

charts/base-cluster/templates/monitoring/kube-prometheus-stack/_prometheus_config.yaml
cwrau: Whoopsie, forgot the `not`

tasches (approved):
reviewed and templated. Looks good to me

PR #546, opened 2023-09-01
feat(base-cluster): remove unused helmrepositories

cwrau:


tasches (approved):
wiremind was used to install the kube-prometheus=stack-crds and is no longer needed. estafette was replaced by trivy.

PR #557, opened 2023-09-08
feat(base-cluster/tracing): simplify tracingConfig

cwrau:
this basically doesn't change anything, as the
open-telemetry-collector-opentelemetry-collector service' internalTrafficPolicy is local anyway


tasches (approved):
tested. Changes were applied correctly.

PR #559, opened 2023-09-11
feat(t8s-cluster)!: change `proxy` to `mirror`

cwrau:
As now we're not really proxying anymore, we're just mirroring.

This means that the nodes are allowed to pull directly from upstream if something is wrong with the mirror.

### DIFF
--- a/charts/t8s-cluster/templates/management-cluster/clusterClass/kubeadmConfigTemplate/_helpers.tpl
+++ b/charts/t8s-cluster/templates/management-cluster/clusterClass/kubeadmConfigTemplate/_helpers.tpl
@@ -2,7 +2,7 @@
 [plugins]
   [plugins."io.containerd.grpc.v1.cri"]
   {{- $_ := set . "Values" .context.Values -}}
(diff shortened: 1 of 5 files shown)


charts/t8s-cluster/templates/management-cluster/clusterClass/kubeadmConfigTemplate/_helpers.tpl
```diff
@@ -38,22 +38,22 @@
     "registry.opensource.zalan.do"
     "registry.teuto.io"
     -}}
-  {{- $proxiedRegistries := concat $defaultProxiedRegistries (.Values.containerRegistryProxy.additionallyProxiedRegistries | default list) | sortAlpha | uniq -}}
-  {{- range $registry := $proxiedRegistries }}
+  {{- $mirroredRegistries := concat $defaultMirroredRegistries (.Values.containerRegistryMirror.additionallyMirroredRegistries | default list) | sortAlpha | uniq -}}
```
marvinWolff, line 41: This could be a breaking change, or am i wrong?

marvinWolff (approved):

PR #578, opened 2023-09-14
feat(base-cluster/ingress): enable tracing for nginx

cwrau:
chore(base-cluster/tracing): simplify tracing config even more


tasches (approved):
tested and reviewed. You can now enable and disable metrics export to opentelemetry.

Issue #487, opened 2023-08-02
[base-cluster] create velero backup monitoring rule


marvinWolff:
We should have a prom-rule to monitor backup success of velero.

tasches:
#724

PR #595, opened 2023-09-28
fix(backup): make it possible to create backups for velero

marvinWolff:
Ref.: https://github.com/teutonet/teutonet-helm-charts/issues/487

### DIFF
--- a/charts/base-cluster/templates/backup/velero.yaml
+++ b/charts/base-cluster/templates/backup/velero.yaml
@@ -91,6 +91,27 @@ spec:
       {{- end }}
       volumeSnapshotLocation: []
     metrics:
+      prometheusRule:
+        additionalLabels: {{- .Values.monitoring.labels | toYaml | nindent 10 }}
+        enabled: true
+        spec:
+        - alert: VeleroBackupPartialFailures
+          annotations:
+            message: Velero backup {{ $labels.schedule }} has {{ $value | humanizePercentage }} partialy failed backups.
+          expr: velero_backup_partial_failure_total{schedule!=""} / velero_backup_attempt_total{schedule!=""} > 0.25
+          for: 15m
+          labels:
+            severity: warning
+        - alert: VeleroBackupFailures
+          annotations:
+            message: Velero backup {{ $labels.schedule }} has {{ $value | humanizePercentage }} failed backups.
+          expr: velero_backup_failure_total{schedule!=""} / velero_backup_attempt_total{schedule!=""} > 0.25
+          for: 15m
+          labels:
+            severity: warning
       serviceMonitor:
+      {{- if .Values.monitoring.prometheus.enabled }}
+        enabled: true
+      {{- end}}
         additionalLabels: {{- .Values.monitoring.labels | toYaml | nindent 10 }}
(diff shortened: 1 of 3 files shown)


charts/base-cluster/templates/backup/velero.yaml
```diff
@@ -91,6 +91,27 @@ spec:
       {{- end }}
       volumeSnapshotLocation: []
     metrics:
+      prometheusRule:
+        additionalLabels: {{- .Values.monitoring.labels | toYaml | nindent 10 }}
+        enabled: true
```
cwrau, line 96: This should also only be enabled if prometheus is activated, just like with the ServiceMonitor, so I would just not set this, as this is also autodetected

charts/base-cluster/values.schema.json
```diff
@@ -1046,6 +1046,11 @@
         "enabled": {
           "type": "boolean"
         },
+        "alerting":{
+          "enabled":{
+            "type": "boolean"
```
cwrau, line 1051: You added this boolean but you don't use it?

I don't get why this boolean exist in any case?

charts/base-cluster/templates/backup/velero.yaml
```diff
@@ -91,6 +91,27 @@ spec:
...
+          annotations:
+            message: Velero backup {{ $labels.schedule }} has {{ $value | humanizePercentage }} failed backups.
+          expr: velero_backup_failure_total{schedule!=""} / velero_backup_attempt_total{schedule!=""} > 0.25
+          for: 15m
+          labels:
+            severity: warning
       serviceMonitor:
+      {{- if .Values.monitoring.prometheus.enabled }}
+        enabled: true
+      {{- end}}
```
cwrau, line 115: This is actually not needed, as this is autodetected by default

cwrau (changes requested):

charts/base-cluster/templates/backup/velero.yaml
marvinWolff: Actually this is needed.

Resolved in call.

cwrau: Nvm, they are stupid

charts/base-cluster/templates/backup/velero.yaml
```diff
@@ -91,6 +91,27 @@ spec:
...
       volumeSnapshotLocation: []
     metrics:
+      prometheusRule:
+        additionalLabels: {{- .Values.monitoring.labels | toYaml | nindent 10 }}
+        enabled: true
+        spec:
+        - alert: VeleroBackupPartialFailures
+          annotations:
+            message: {{ `Velero backup {{ $labels.schedule }} has {{ $value | humanizePercentage }} partialy failed backups.` }}
+          expr: velero_backup_partial_failure_total{schedule!=""} / velero_backup_attempt_total{schedule!=""} > 0.25
```
cwrau, line 101: Isn't there a nicer way to really detect backup failure?

These metrics only point to problems, as with the current one we could have 3 failed attempts and 1 successful one (which is good in my eyes) which would still trigger the alert

PR #714, opened 2024-01-12
feat(base-cluster): add weaveworks gitops dashboard chart

### LINKED ISSUE #384: [base-cluster] maybe add weaveworks dashboard
See https://github.com/weaveworks/weave-gitops/tree/main/charts/gitops-server

Nice UI, non-invasive and doesn't break gitops 😁

tasches:
closes #384


charts/base-cluster/values.schema.json
```diff
@@ -1263,6 +1263,102 @@
...
+            "username": {
+              "type": "string"
+            },
+            "createSecret": {
+              "type": "boolean"
+            },
+            "createClusterRole": {
+              "type": "boolean"
+            },
+            "passwordHash": {
```
cwrau, line 1288: This should be the plain password and `bcrypt`ed via helm

charts/base-cluster/values.schema.json
```diff
@@ -1263,6 +1263,102 @@
...
+            },
+            "host": {
+              "type": "string"
+            },
+            "tls": {
+              "type": "array"
+            },
+            "className": {
+              "type": "string"
+            }
```
cwrau, line 1307: The user doesn't care about this

charts/base-cluster/values.schema.json
```diff
@@ -1263,6 +1263,102 @@
...
+        "oidcSecret": {
+          "type": "object",
+          "properties": {
+            "additionalKeys": {
+              "type": "object"
+            },
+            "create": {
+              "type": "boolean"
+            }
+          }
```
cwrau, line 1319: This should be handled like grafana, using `global.authentication`

charts/base-cluster/values.schema.json
```diff
@@ -1263,6 +1263,102 @@
...
+          "impersonationResourceNames": {
+            "type": "array"
+          },
+          "impersonationResources": {
+            "type": "array"
+          },
+          "viewSecretsResourceNames": {
+            "type": "array"
+          }
+        },
```
cwrau, line 1340: What is this used for? We should provide a common option or just leave this out

I'm in favor of the latter

charts/base-cluster/values.schema.json
```diff
@@ -1263,6 +1263,102 @@
...
+                  "type": "object"
+                },
+                "port": {
+                  "type": "integer"
+                }
+              }
+            }
+          }
+        }
+      }
```
cwrau, line 1360: Should the user be able to turn this off?

I don't think so

charts/base-cluster/templates/weave/gitops.yaml
cwrau, line 1: I would rename this to `gitops-dashboard.yaml`

charts/base-cluster/templates/weave/gitops.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+  chart:
+    spec:
+      chart: weave-gitops
+      sourceRef:
+        kind: HelmRepository
+        name: weaveworks
+  interval: 5m
+  values:
+  {{- .Values.weaveworks | toYaml | nindent 4 }}
```
cwrau, line 19: Definitely not like this, we should provide a nice, usable DSL, not just let the user write the plain values, see other comments

charts/base-cluster/values.schema.json
tasches: unfortunately this is not supported by the chart. You can enter a password hash, which is generated by their cli tool :disappointed:

charts/base-cluster/templates/weave/gitops.yaml
tasches: I just want to limit the values which the user can enter with the values schema but keep the configuration abillity the gitops-chart offers

charts/base-cluster/values.schema.json
tasches: if the user is not interessted in metrics from the dashboard he can turn it off. However, we can drop the service section because this isn"t necessary in most cases

charts/base-cluster/values.schema.json
tasches: than lets drop it. The default values should to the job perfectly fine :+1:

charts/base-cluster/values.schema.json
cwrau: >if the user is not interessted in metrics from the dashboard he can turn it off

Yeah, but will that ever happen? 🤷

charts/base-cluster/values.schema.json
cwrau: It should be, why not?

charts/base-cluster/values.schema.json
```diff
@@ -1263,6 +1263,102 @@
...
+            },
+            "username": {
+              "type": "string"
+            },
+            "createSecret": {
+              "type": "boolean"
+            },
+            "createClusterRole": {
+              "type": "boolean"
+            },
```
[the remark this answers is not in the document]
tasches: if the default password should be used, the createSecret is not necessary. For the createClusterRole we should keep the chart default

charts/base-cluster/templates/weave/gitops.yaml
cwrau: I understand, but we should make it better than the chart

Even just having a plain password instead of having to hash it yourself is worth it

charts/base-cluster/values.schema.json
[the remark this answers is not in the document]
cwrau: > if the default password should be used, the createSecret is not necessary

Yeah, but A) we can decide this internally, B) even if it's not necessary, it doesn't hurt and would reduce configuration complexity and C) we should provide a generated default password and not use the default one

charts/base-cluster/values.schema.json
tasches: removed

charts/base-cluster/templates/weave/gitops.yaml
tasches: reformated

charts/base-cluster/templates/weave/gitops.yaml
tasches: renamed

charts/base-cluster/values.schema.json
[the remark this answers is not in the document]
tasches: I changed it to if adminUser enabled than create the secret. It make sense to me :grin:

charts/base-cluster/values.schema.json
tasches: can you take a look at the current state. Haven"t tested it though

charts/base-cluster/values.schema.json
tasches: This is also not tested yet but I generated the passwordHash with the httpwd helm function because the chat expected a bcrypt hash as input

tasches: I did some adjustsments after the test. However, now it works. Maybe I should move the calculation to the _helper.tpl. What do you think?

charts/base-cluster/templates/NOTES.txt
```diff
@@ -40,6 +40,24 @@ Password: $ kubectl -n monitoring get secret {{ $secretName | quote }} -o json |
 {{- end }}
 {{- end }}
 
+{{- if .Values.weaveworks.enabled }}
+
+===
+You can enter your gitops-dashboard via
+{{- if and .Values.weaveworks.ingress .Values.weaveworks.ingress.enabled }}
+{{- printf "https://%s" .Values.weaveworks.ingress.host | nindent 2 }}
```
cwrau, line 48: This is not the full host

charts/base-cluster/templates/weave/gitops-dashboard.yaml
```diff
@@ -0,0 +1,56 @@
+{{- if .Values.weaveworks.enabled }}
+apiVersion: v1
```
cwrau, line 2: I would create the namespace via the `global.namespaces` feature

charts/base-cluster/templates/weave/gitops-dashboard.yaml
```diff
@@ -0,0 +1,56 @@
...
+kind: HelmRelease
+metadata:
+  annotations:
+    metadata.weave.works/description: This is the Weave GitOps Dashboard.  It provides
+      a simple way to get insights into your GitOps workloads.
+  name: gitops-dashboard
+  namespace: weave
+spec:
+  chart:
+    spec:
```
cwrau, line 18: There is a template function for creating this block, see other HelmReleases

charts/base-cluster/templates/weave/gitops-dashboard.yaml
```diff
@@ -0,0 +1,56 @@
...
+  interval: 5m
+  values:
+    replicaCount: 1
+    {{- if and .Values.weaveworks.adminUser .Values.weaveworks.adminUser.create }}
+    adminUser:
+      create: true
+      username: {{ .Values.weaveworks.adminUser.username }}
+      createSecret: true
+      {{- if .Values.weaveworks.adminUser.password }}
+      {{- $password := split ":" (htpasswd .Values.weaveworks.adminUser.username .Values.weaveworks.adminUser.password) }}
```
cwrau, line 33: Shouldn't this be bcrypt? 🤔

charts/base-cluster/templates/weave/gitops-dashboard.yaml
```diff
@@ -0,0 +1,56 @@
...
+      passwordHash: {{  $password._1 }}
+      {{- end }}
+    {{- end }}
+    {{- if and .Values.weaveworks.ingress .Values.weaveworks.ingress.enabled }}
+    ingress:
+      enabled: true
+      host: {{ .Values.weaveworks.ingress.host }}
```
cwrau, line 40: This is not the full ingress url, see other ingress', like grafana

charts/base-cluster/values.schema.json
```diff
@@ -1263,6 +1263,81 @@
...
+      "then": {
+        "anyOf": [
+          {
+            "required": ["adminUser"]
+          },
+          {
+            "required": ["oidcSecret"]
+          }
+        ]
+      },
```
cwrau, line 1338: Instead of `if`, `then`, ... I'd rather just use `anyOf`

In this case this isn't even needed, as I'd set the adminUser by default to `admin` and auto-generate a password if none is set, just like for the grafana admin user

charts/base-cluster/values.schema.json
```diff
@@ -1263,6 +1263,81 @@
...
+      "type": "object",
+      "description": "Values for the gitops-chart",
+      "properties": {
+        "enabled": {
+          "type": "boolean"
+        },
+        "adminUser": {
+          "type": "object",
+          "properties": {
+            "create": {
```
cwrau, line 1276: I would drop this `create` flag, the user should always be created, with default `admin` name and a generated password

charts/base-cluster/values.schema.json
```diff
@@ -1263,6 +1263,81 @@
...
+            "host": {
+              "type": "string"
+            }
+          },
+          "additionalProperties": false
+        },
+        "oidcSecret": {
+          "type": "object",
+          "properties": {
+            "create": {
```
cwrau, line 1314: Kinda the same thing here, there is no need for a `create` parameter, we will connect the OIDC if `global.authentication` is configured, just like with grafana

cwrau (changes requested):

charts/base-cluster/templates/weave/gitops-dashboard.yaml
tasches: replaced. Looked up the function in the helm docu.

charts/base-cluster/values.schema.json
tasches: Should I remove the possibility for the user to set different credentials?

charts/base-cluster/values.schema.json
tasches: I read a post about helm validations and json schema on stackoverflow. It was recommend to use if then instead of anyOf because it is easier to read, while the same functionality can be kept in that case

cwrau: Mh, I personally definitely don't think that it's easier to read randomly ordered json properties than plain `anyOf`

But anyways, you're still using `anyOf` *and* neither is needed here

charts/base-cluster/values.schema.json
cwrau: What do you mean by "different credentials"?

The user can supply a password and/or provide OIDC configuration. Is there something else?

charts/base-cluster/values.schema.json
tasches: oidc-configuration removed. It will be enabled, if global.authentication is enabled

charts/base-cluster/values.schema.json
tasches: got it. That was the question ;)

charts/base-cluster/templates/weave/gitops-dashboard.yaml
tasches: refactored

charts/base-cluster/templates/weave/_helpers.tpl
```diff
@@ -0,0 +1,3 @@
+{{- define "weaveworks.password" -}}
+{{ include "common.secrets.passwords.manage" (dict "secret" "cluster-user-auth" "key" "password" "providedValues" (list "weaveworks.adminUser.password") "length" 20 "strong" false "skipB64enc" true "skipQuote" true "context" $) | bcrypt }}
+{{- end -}}
```
cwrau, line 3: This is only used once, please inline

charts/base-cluster/templates/weave/_helpers.tpl
```diff
@@ -0,0 +1,3 @@
+{{- define "weaveworks.password" -}}
+{{ include "common.secrets.passwords.manage" (dict "secret" "cluster-user-auth" "key" "password" "providedValues" (list "weaveworks.adminUser.password") "length" 20 "strong" false "skipB64enc" true "skipQuote" true "context" $) | bcrypt }}
```
cwrau, line 2: Do we *really* need bcrypt? Then you have to totally revamp this, as bcypt is random every time, see https://gitlab.teuto.net/4teuto/dev/teuto-portal/teuto-portal/-/blob/main/charts/teuto-portal/templates/mailpit/auth-secret.yaml

We can also talk about this, as it's kinda complicated

charts/base-cluster/values.yaml
```diff
@@ -43,6 +43,8 @@ global:
       condition: "{{ .Values.backup.enabled }}"
       additionalLabels:
         app.kubernetes.io/component: backup
+    weave:
+      condition: "{{ .Values.weaveworks.enabled }}"
```
cwrau, line 47: Mh, really into it's own namespace?

I would rather put this in flux-system 🤔

charts/base-cluster/values.yaml
```diff
@@ -444,3 +451,6 @@ backup:
       limits:
         cpu: 500m
         memory: 500Mi
+
+weaveworks:
+  enabled: false
```
cwrau, line 456: Depending on the resource usage we could also deploy it by default, what do you think? @tasches @marvinWolff

cwrau (changes requested):

charts/base-cluster/templates/NOTES.txt
```diff
@@ -40,6 +40,24 @@ Password: $ kubectl -n monitoring get secret {{ $secretName | quote }} -o json |
...
+{{- if and .Values.weaveworks.ingress .Values.weaveworks.ingress.enabled }}
+{{- printf "https://%s" (include "base-cluster.weaveworks.host" $) | nindent 2 }}
+{{- else }}
+{{- printf "$ kubectl -n monitoring port-forward svc/gitops-dashboard 9901:http" | nindent 2 }}
+{{- end }}
+{{- if and .Values.weaveworks.oidcSecret .Values.weaveworks.oidcSecret.create }}
+Login via your OIDC provider
+{{- else}}
+{{- if .Values.weaveworks.adminUser }}
+{{- printf "username: %s\npassword: %s" .Values.weaveworks.adminUser.username .Values.weaveworks.adminUser.password | nindent 2}}
```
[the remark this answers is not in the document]
tasches: Yeah I know. This is the part I wanted to update anyways. Because of the bcrypt function this won't work at all

charts/base-cluster/ci/artifacthub-values.yaml
```diff
@@ -27,3 +27,11 @@ monitoring:
...
           enabled: true
           integrationKey: INTEGRATION_KEY
+weaveworks:
+  enabled: true
+  adminUser:
+    username: admin
+    password: test
+  ingress:
+    enabled: true
+    customDomain: ww-dashboard.local
```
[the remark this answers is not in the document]
tasches: I just wanted to show a possibe configuration with an ingress to the dashboard. I guess most of the time you want access to the dashboard without the struggle to use a portforward all the time, right?

charts/base-cluster/templates/weave/_helpers.tpl
tasches: well this must be changed anywayse. As you also mentioned in the comment below. This won't work

charts/base-cluster/values.yaml
tasches: done

charts/base-cluster/templates/weave/_helpers.tpl
tasches: unfortunately, yes. The dashboard expects a bcrypted input and there is no other option to deploy a password for the user.

charts/base-cluster/ci/artifacthub-values.yaml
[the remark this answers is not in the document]
cwrau: Examples should be in the `ci` folder yamls

Here should only be real default values that make sense

cwrau (changes requested):

charts/base-cluster/values.yaml
```diff
@@ -458,3 +463,8 @@ backup:
...
+  adminUser:
+    username: admin
```
tasches, line 470: since you can only configure one user the path should change from `adminUser.username `and `admiUser.password` to just username, password

charts/base-cluster/templates/weave/credentials.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: Secret
+metadata:
+  name: cluster-user-auth
+  namespace: flux-system
+  labels: {{- include "common.labels.standard" $ | nindent 4 }}
+    app.kubernetes.io/component: gitops-dashboard
+stringData:
+  password: {{ $pwd | bcrypt | quote }}
+  username: {{ .Values.weaveworks.adminUser.username | quote }}
+  password-dec: {{ $pwd | quote }}
```
tasches, line 14: Currently you can not update the password afterwards. If you change the password in the values file after an initial installation, it won't update

charts/base-cluster/templates/NOTES.txt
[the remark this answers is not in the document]
tasches: updated

charts/base-cluster/templates/weave/_helpers.tpl
tasches: updated

charts/base-cluster/values.yaml
tasches: It does not require a lot of the resources. So we can enabled it per default. Should I leave the enable flag? So the user have the possibility to disable the dashboard if he wants to?

charts/base-cluster/templates/weave/_helpers.tpl
tasches: deleted

charts/base-cluster/ci/artifacthub-values.yaml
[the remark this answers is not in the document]
tasches: removed

charts/base-cluster/templates/weave/credentials.yaml
tasches: fixed but the password which is printed in the notes is now wrong

charts/base-cluster/templates/weave/gitops-dashboard.yaml
```diff
@@ -0,0 +1,38 @@
...
+    replicaCount: 1
+    {{- if and .Values.weaveworks.ingress .Values.weaveworks.ingress.enabled }}
+    ingress:
+      enabled: true
+      host: {{ include "base-cluster.weaveworks.host" $ }}
+    {{- end }}
+    metrics:
+      enabled: {{ .Values.monitoring.prometheus.enabled }}
+    {{- with .Values.global.authentication }}
+    {{- if and .config.clientId $.Values.weaveworks.ingress $.Values.weaveworks.ingress.host }}
```
tasches, line 24: @cwrau not sure if the redirectURL is actual mandatory for the OIDC :thinking:. However, due to ci issues I disabled the oidc-creation if the ingress is not configured for the gitops-dashboard. We can also drop the redirectURL and as the result the condition as well, if we do not need the redirectURL at all

tasches:
converted to draft because the weavedashboard is outdated. There is a refactoring of the dashboard in progress.

PR #724, opened 2024-01-17
feat(base-cluster): enable velero servicemonitor if prometheus is enabled

### LINKED ISSUE #487: [base-cluster] create velero backup monitoring rule

tasches:
closes #487

### DIFF
--- a/charts/base-cluster/templates/backup/velero.yaml
+++ b/charts/base-cluster/templates/backup/velero.yaml
@@ -100,10 +100,10 @@ spec:
         additionalLabels: {{- .Values.monitoring.labels | toYaml | nindent 10 }}
         spec:
          - alert: VeleroBackupFailures
-            annotations:
-              message: Velero backup {{ "{{ $labels.schedule }}" }} failed.
-            expr: velero_backup_last_status{schedule!=""}!=1
-            for: 15m
-            labels:
-              severity: warning
+           annotations:
+             message: Velero backup {{ "{{ $labels.schedule }}" }} failed.
+           expr: velero_backup_last_status{schedule!=""}!=1
+           for: 15m
+           labels:
+             severity: warning
 {{- end }}


charts/base-cluster/templates/backup/velero.yaml
```diff
@@ -92,5 +92,8 @@ spec:
       volumeSnapshotLocation: []
     metrics:
       serviceMonitor:
+        {{- if .Values.monitoring.prometheus.enabled }}
```
cwrau, line 95: The `if` is not needed, we can just always just enable the serviceMonitor, as the chart checks for the existence of the CRD

cwrau (changes requested):

marvinWolff:
This does enable the metrics endpoints, but will the alerts also be generated?  
As the endpoints alone will not give us any alerts if a backup fails.

marvinWolff:
https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/values.yaml#L261

This needs to be set in order for the rules to be created.

charts/base-cluster/templates/backup/velero.yaml
tasches: removed

charts/base-cluster/templates/backup/velero.yaml
```diff
@@ -87,5 +87,23 @@ spec:
       volumeSnapshotLocation: []
     metrics:
       serviceMonitor:
+        enabled: true
```
cwrau, line 90: This should be conditional to prometheus being enabled, or are they smart and auto-detect prometheus?

Also, you should add a `dependsOn` for promtheus, if it's enabled

charts/base-cluster/templates/backup/velero.yaml
```diff
@@ -87,5 +87,23 @@ spec:
...
         additionalLabels: {{- .Values.monitoring.labels | toYaml | nindent 10 }}
+      prometheusRule:
+        enabled: true
```
cwrau, line 93: Same thing here

charts/base-cluster/templates/backup/velero.yaml
```diff
@@ -87,5 +87,23 @@ spec:
...
+            for: 15m
+            labels:
+              severity: warning
+          - alert: VeleroBackupFailures
+            annotations:
+              message: Velero backup {{ "{{ $labels.schedule }}" }} has {{ "{{ $value | humanizePercentage }} "}} failed backups.
+            expr: velero_backup_failure_total{schedule!=""} / velero_backup_attempt_total{schedule!=""} > 0.25
+            for: 15m
+            labels:
+              severity: warning
```
cwrau, line 108: We should overthink this, these alarms are fading over time

We should use the metric `velero_backup_last_status`, if it's working now (in the old version the status wasn't always correct... f*ck velero 🤮)

cwrau (dismissed):

charts/base-cluster/templates/backup/velero.yaml
tasches: I removed the check because you mentioned in an earlier comment, that the velero chart performs the check on its own

charts/base-cluster/templates/backup/velero.yaml
tasches: I just copied it from the chart values.yaml :sweat_smile: 
So I guess it does not work as expected :rofl:

charts/base-cluster/templates/backup/velero.yaml
cwrau: Ah, oki dokes 👍

cwrau: But the `dependsOn` should be added nonetheless

tasches:
screenshot from a test deployment. The PrometheusRules are now visible an active
image

charts/base-cluster/templates/backup/velero.yaml
cwrau: Yes

tasches: Just for the documentation 
https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/templates/prometheusrule.yaml#L1

charts/base-cluster/templates/backup/velero.yaml
marvinWolff: Do we even want to deploy rules?

Everyone could need a different set of rules, so i'm not sure if we want this.

cwrau: > Do we even want to deploy rules?

Definitely, otherwise monitoring this is useless

> Everyone could need a different set of rules, so i'm not sure if we want this.

These rules will be generic if we use `velero_backup_last_status`, therefore they work for everyone

The only thing we can't monitor is, that backups are attempted when they should, but that's a currently unsolvable problem, like with "alert if a desired cluster is missing"..., as we don't have metrics for "should exist"

tasches: I have implemented the requested changes

marvinWolff (approved):

PR #725, opened 2024-01-17
chore(ci): migrate release-please action

cwrau:
hopefully this works 🤞


.github/release-please/config.json
```diff
@@ -0,0 +1,37 @@
...
+    "charts/stellio-context-broker": {
+      "package-name": "stellio-context-broker"
+    },
+    "charts/t8s-cluster": {
+      "package-name": "t8s-cluster"
+    },
+    "charts/teuto-portal-k8s-worker": {
+      "package-name": "teuto-portal-k8s-worker"
+    }
+  }
```
cwrau, line 36: I know this is sad, but I couldn't find anything about specyfing something like "package regex" or "just use all packages with their foldernames as names" and the ability to just say "do that package" has been dropped 😥

PR #763, opened 2024-02-05
fix(base-cluster/backup)!: only enable velero if a backupstoragelocation is set

### LINKED ISSUE #752: [base-cluster] only install velero when a default storage location is defined
Otherwise velero won't start. 

I'm so glad that we use velero.

tasches:
closes #752

### DIFF
--- a/charts/base-cluster/templates/backup/validation.yaml
+++ b/charts/base-cluster/templates/backup/validation.yaml
@@ -1,4 +1,4 @@
-{{- if and .Values.backup.enabled .Values.backup.defaultLocation -}}
+{{- if .Values.backup.defaultLocation -}}
   {{- if not (index .Values.backup.backupStorageLocations .Values.backup.defaultLocation) -}}
     {{- fail (printf "The `defaultLocation` '%s' must exist in `backupStorageLocations`" .Values.backup.defaultLocation) -}}
   {{- end -}}
--- a/charts/base-cluster/values.yaml
+++ b/charts/base-cluster/values.yaml
@@ -166,7 +166,7 @@ global:
       url: https://vmware-tanzu.github.io/helm-charts
       charts:
         velero: 5.x
-      condition: "{{ .Values.backup.backupStorageLocations }}"
+      condition: "{{ not (empty .Values.backup.backupStorageLocations) }}"
     open-telemetry:
       url: https://open-telemetry.github.io/opentelemetry-helm-charts
       charts:
--- a/charts/base-cluster/values.yaml
+++ b/charts/base-cluster/values.yaml
@@ -56,7 +56,7 @@ global:
         app.kubernetes.io/component: storage
         app.kubernetes.io/part-of: nfs-server-provisioner
     backup:
-      condition: "{{ .Values.backup.enabled }}"
+      condition: "{{ .Values.backup.backupStorageLocations }}"
       additionalLabels:
         app.kubernetes.io/component: backup
   certificates:
--- a/charts/base-cluster/values.yaml
+++ b/charts/base-cluster/values.yaml
@@ -166,7 +166,7 @@ global:
       url: https://vmware-tanzu.github.io/helm-charts
       charts:
         velero: 5.x
-      condition: "{{ .Values.backup.enabled }}"
+      condition: "{{ .Values.backup.backupStorageLocations }}"
     open-telemetry:
       url: https://open-telemetry.github.io/opentelemetry-helm-charts
       charts:
@@ -443,7 +443,7 @@ rbac:
   accounts: {}
 
 backup:
-  enabled: true
+  enabled: false
   backupStorageLocations: {}
   resources:
     requests:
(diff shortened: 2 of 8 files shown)


marvinWolff:
This looks fine to me.

But i'm asking myself how the rollout will go.

Most clusters currently aren't using velero, but some do. In those clusters the backupStorageLocation ist not set via the chart. At least not that i'm aware of.  
Here we should either move the declaration into the chart or add some sort of configuration option ala `existingStorageLocation: <name>`.

But im not sure how practical that would be. What do you think @cwrau @tasches ?

charts/base-cluster/templates/backup/validation.yaml
```diff
@@ -3,3 +3,9 @@
     {{- fail (printf "The `defaultLocation` '%s' must exist in `backupStorageLocations`" .Values.backup.defaultLocation) -}}
   {{- end -}}
 {{- end -}}
+{{- if .Values.backup.enabled -}}
+  {{- $length := len .Values.backup.backupStorageLocations -}}
+  {{- if eq $length 0 -}}
+    {{- fail (print "At least one .Values.backup.backupstoragelocations needs to be configured")  -}}
+  {{- end -}}
+{{- end -}}
```
cwrau, line 11: You should replace this whole block with the following inside the schema;

```diff
diff --git a/charts/base-cluster/values.schema.json b/charts/base-cluster/values.schema.json
index 18c64f5..62b367b 100644
--- a/charts/base-cluster/values.schema.json
+++ b/charts/base-cluster/values.schema.json
@@ -1278,7 +1278,8 @@
               "bucket"
             ],
             "additionalProperties": false
-          }
+          },
+          "minProperties": 1
         },
         "defaultLocation": {
           "type": "string"
@@ -1293,6 +1294,9 @@
           "additionalProperties": false
         }
       },
+      "required": [
+        "defaultLocation"
+      ],
       "additionalProperties": false
     },
     "common": {
```

which should accomplish the same

charts/base-cluster/values.yaml
```diff
@@ -443,7 +443,7 @@ rbac:
   accounts: {}
 
 backup:
-  enabled: true
+  enabled: false
```
cwrau, line 446: This is a breaking change, which is fine, you just need to change your commit message to include an `!`; `fix(base-cluster)!:`

And you should include the subarea in you message; `fix(base-cluster/backup)!:`

cwrau (dismissed):

charts/base-cluster/templates/backup/validation.yaml
tasches: no longer necessary

cwrau:
You're missing some locations where `.backup.enabled` is used;

```
charts/base-cluster/values.yaml
charts/base-cluster/templates/backup/velero.yaml
charts/base-cluster/templates/backup/validation.yaml
charts/base-cluster/templates/backup/migrations/velero-4-to-5.yaml
charts/base-cluster/templates/monitoring/kube-prometheus-stack/_grafana-config.yaml
```

You can just search for `.backup.enabled`

cwrau:
You should require `defaultLocation` if `backupStorageLocations` is defined;

```diff
diff --git a/charts/base-cluster/values.schema.json b/charts/base-cluster/values.schema.json
index d8f2f69..948cc3b 100644
--- a/charts/base-cluster/values.schema.json
+++ b/charts/base-cluster/values.schema.json
@@ -1290,6 +1290,25 @@
           "additionalProperties": false
         }
       },
+      "oneOf": [
+        {
+          "properties": {
+            "backupStorageLocations": {
+              "minProperties": 1
+            }
+          },
+          "required": [
+            "defaultLocation"
+          ]
+        },
+        {
+          "properties": {
+            "backupStorageLocations": {
+              "maxProperties": 0
+            }
+          }
+        }
+      ],
       "additionalProperties": false
     },
     "common": {
```

In this case I'd also be fine with the `if`/`then`;

```diff
diff --git a/charts/base-cluster/values.schema.json b/charts/base-cluster/values.schema.json
index d8f2f69..4ac7ee7 100644
--- a/charts/base-cluster/values.schema.json
+++ b/charts/base-cluster/values.schema.json
@@ -1290,6 +1290,18 @@
           "additionalProperties": false
         }
       },
+      "if": {
+        "properties": {
+          "backupStorageLocations": {
+            "minProperties": 1
+          }
+        }
+      },
+      "then": {
+        "required": [
+          "defaultLocation"
+        ]
+      },
       "additionalProperties": false
     },
     "common": {
```

charts/base-cluster/values.yaml
```diff
@@ -56,7 +56,7 @@ global:
         app.kubernetes.io/component: storage
         app.kubernetes.io/part-of: nfs-server-provisioner
     backup:
-      condition: "{{ .Values.backup.enabled }}"
+      condition: "{{ not (empty .Values.backup.backupStorageLocations) }}"
```
marvinWolff, line 59: Is this `not empty` really needed?

I'm unsure. We haven't used it anywhere else, but i think it also won't hurt?

marvinWolff: Alright, was cleared. It's ok

marvinWolff (approved):

PR #784, opened 2024-02-13
chore(base-cluster/kyverno)!: upgrade 🤮

cwrau:


charts/base-cluster/values.yaml
```diff
@@ -137,9 +137,8 @@ global:
     kyverno:
       url: https://kyverno.github.io/kyverno
       charts:
-        # TODO: migrate to v3 https://artifacthub.io/packages/helm/kyverno/kyverno#migrating-from-v2-to-v3
-        kyverno: 2.x.x
-        kyverno-policies: 2.x.x
+        kyverno: 3.x.x
```
tasches, line 140: regarding the kyverno documentation this is a breaking change, which is not possible right away. I see you mentioned this in the chart validation but you should write a small note in the README template as well :wink:

tasches (changes requested):

tasches (approved):
tested and reviewed

PR #791, opened 2024-02-16
fix(base-cluster-schema/schema): remove required for serviceLevelAgre…

cwrau:
…ement

the `enum` directive already forces this values to be specified but this way the schema validates if you don't explicitly specify it


tasches (approved):
fixed PR name :grin: . Rest looks good

PR #792, opened 2024-02-16
feat: add helmrelease schemas 😍

cwrau:
That way we can use 

```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/teutonet/teutonet-helm-charts/main/charts/base-cluster/helmrelease.schema.json
```

in the `cluster.yaml`s and get correct validation, completion, ...


tasches (approved):
I like it :+1:

cwrau:
I only tested this with intellij, which worked because it ignored some errors I guess and just pulled the schema for the HelmRelease from the cluster

See https://github.com/fluxcd/flux2/discussions/408#discussioncomment-8492078

PR #835, opened 2024-03-26
feat(base-cluster): overhaul resources

cwrau:
chore(base-cluster): provide better defaults for resources


charts/base-cluster/values.yaml
tasches, line 1: good idea but were these limits tested?

cwrau: Jein, I compared them to the existing ones and if they weren't compatible I looked at clusters to see if that's ok otherwise they would be ok anyways

PR #854, opened 2024-04-11
feat(common): initial implementation

cwrau:
adjust CI to work with library charts

### DIFF
--- a/.github/scripts/create-values-diff.sh
+++ b/.github/scripts/create-values-diff.sh
@@ -22,6 +22,11 @@ else
   dryRun=false
 fi
 
+if yq -e '.type == "library"' "$chart/Chart.yaml" >/dev/null; then
+  echo "Skipping library chart '$chart'" >/dev/stderr
+  exit 0
+fi
+
 GITHUB_API_URL="${GITHUB_API_URL:-https://api.github.com}"
 
 if command -v gh &>/dev/null; then
@@ -45,7 +50,7 @@ trap 'rm -rf "$TMP_DIR"' EXIT
 cd "$GITHUB_WORKSPACE"
 
 function generateComment() {
-  local chart="charts/${1?}"
+  local chart="${1?}"
   local -A diffs
   local newResourcesDir
   local originalResourcesDir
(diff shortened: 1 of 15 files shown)


.github/scripts/create-values-diff.sh
```diff
@@ -45,7 +50,7 @@ trap 'rm -rf "$TMP_DIR"' EXIT
 cd "$GITHUB_WORKSPACE"
 
 function generateComment() {
-  local chart="charts/${1?}"
+  local chart="${1?}"
```
tasches, line 53: I cannot see why the path changed from charts/$ {1} to just {$1}

marvinWolff: Because he changed the way it's called.

You can see this here: https://github.com/teutonet/teutonet-helm-charts/pull/854/files#diff-1c060ea3361efbeee5f55dc7e6086ed9cdf182099274e382d630cac8a092dc60R28

marvinWolff: My guess is, he did it to make it more universal. Or is there a different reason @cwrau?

charts/common/templates/_resources.tpl
```diff
@@ -0,0 +1,22 @@
...
+Return a resource request/limit object based on a given preset or provided resources.
+{{ include "common.resources" (dict "resources" (dict) "resourcesPreset" "small") -}}
+*/}}
+{{- define "common.resources" -}}
+  {{- $resources := dict -}}
+  {{- if .resources -}}
+    {{- $resources = .resources | merge $resources -}}
+  {{- end -}}
+  {{- if and .resourcesPreset (ne .resourcesPreset "none") -}}
+    {{- $resources = include "common.resources.preset" (dict "type" .resourcesPreset) | fromYaml | merge $resources -}}
```
marvinWolff, line 11: I don't get where the resources are coming from.  
So, i've seen the sizes in the file below, but how does the resource object actually look like?

marvinWolff: Or is this something the bitnami charts just understand?

marvinWolff: Ok, nvm, found it (my google fu wasn't strong enough). This is genius: https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_resources.tpl#L16

marvinWolff (approved):

tasches (approved):

PR #916, opened 2024-04-29
 feat(ckan): add alpha ckan helmchart

SyeKlu:


charts/ckan/templates/ingress.yaml
```diff
@@ -0,0 +1,29 @@
...
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- if .Values.ckan.ingress.className }}
+  ingressClassName: {{ .Values.ckan.ingress.className }}
+  {{- end }}
+  tls:
+    - hosts:
+        - {{ required "You must provide a host, when enabling ingress" .Values.ckan.ingress.baseDomain | quote }}
+      secretName: {{ include "common.secrets.name" (dict "existingSecret" .Values.ckan.ingress.existingCertificate "defaultNameSuffix" "custom-certificate" "context" $) }}
```
cwrau, line 21: The `defaultNameSuffix` is used when no existingCertificate is given, therefore it should be something like `certificate`

```suggestion
      secretName: {{ include "common.secrets.name" (dict "existingSecret" .Values.ckan.ingress.existingCertificate "defaultNameSuffix" "certificate" "context" $) }}
```

charts/ckan/values.schema.json
```diff
@@ -0,0 +1,259 @@
...
+              "annotations": {
+                "type": "object"
+              },
+              "baseDomain": {
+                "type": "string",
+                "description": "The base domain to be used for cluster ingress",
+                "examples": [
+                  "teuto.net"
+                ]
+              }
```
cwrau, line 58: This is never _used_ as a real base domain, why not `host`?

Especially since this in the `ckan` block, so explicitly *not* a baseDomain

And the description doesn't match 😉

charts/ckan/templates/ingress.yaml
```diff
@@ -0,0 +1,29 @@
...
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- if .Values.ckan.ingress.className }}
+  ingressClassName: {{ .Values.ckan.ingress.className }}
+  {{- end }}
+  tls:
+    - hosts:
+        - {{ required "You must provide a host, when enabling ingress" .Values.ckan.ingress.baseDomain | quote }}
```
cwrau, line 20: This is better handled in the schema

charts/ckan/values.schema.json
```diff
@@ -0,0 +1,259 @@
...
+          "replicas": {
+            "type": "integer"
+          },
+          "siteId":{
+            "type": "string",
+            "decription": ""
+          },
+          "siteTitle": {
+            "type": "string",
+            "description": ""
```
cwrau, line 35: Empty?

charts/ckan/values.schema.json
```diff
@@ -0,0 +1,259 @@
...
+      },
+      "ckan": {
+        "type": "object",
+        "properties": {
+          "replicas": {
+            "type": "integer"
+          },
+          "siteId":{
+            "type": "string",
+            "decription": ""
```
cwrau, line 31: Typo, but empty anyways?

charts/ckan/values.schema.json
```diff
@@ -0,0 +1,259 @@
...
+              "type": "string"
+            },
+            "datapusherPassword": {
+              "type": "string"
+            },
+            "datapusherDb": {
+              "type": "string"
+            }
+          }
+        }
```
cwrau, line 188:
```suggestion
        "ckanDbs": {
          "type": "object",
          "patternProperties": {
            "datastore|datapusher": {
              "type": "object",
              "properties": {
                "username": {
                  "type": "string"
                },
                "password": {
                  "type": "string"
                },
                "db": {
                  "type": "string"
                }
              },
              "additionalProperties": false
            }
          }
        }
```

charts/ckan/values.schema.json
```diff
@@ -0,0 +1,259 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema",
+  "type": "object",
+  "title": "ckan configuration",
+  "properties": {
```
cwrau, line 5: The schema should be more strict, you basically need `additionalProperties: false` on all objects and you should add `required: []` where applicable.

Things like the securityContext should only be configurable if necessary, as the is rarely able to change something here, more often than not, it should just be hard-coded locked-down

charts/ckan/values.schema.json
```diff
@@ -0,0 +1,259 @@
...
+              "mailFrom": {
+                "$ref": "#/$defs/email"
+              },
+              "tls": {
+                "type": "string",
+                "enum": [
+                  "enabled",
+                  "disabled"
+                ]
+              },
```
cwrau, line 119:
```suggestion
              "tls": {
                "type": "object",
                "properties": {
                  "enabled": {
                    "type": "boolean"
                  }
                },
                "additionalProperties": false
              },
```

charts/ckan/templates/ckan/deployment.yaml
```diff
@@ -0,0 +1,115 @@
...
+  replicas: {{ .Values.ckan.replicas }}
+  selector:
+    matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
+      app.kubernetes.io/component: {{ $name }}
+  template:
+    metadata:
+      {{- with .Values.ckan.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
cwrau, line 19: What would this be used for?

charts/ckan/values.schema.json
```diff
@@ -0,0 +1,259 @@
...
+      "redis": {
+        "enabled": {
+          "type": "boolean"
+        }
+      },
+      "solr": {
+        "enabled": {
+          "type": "boolean"
+        }
+      }
```
cwrau, line 199: That's not how that works, missing the `properties` field and `additionalProperties: false` as well

charts/ckan/values.schema.json
```diff
@@ -0,0 +1,259 @@
...
+    }
+  },
+  "$defs": {
+    "id": {
```
cwrau, line 203: This is never used?

charts/ckan/values.yaml
```diff
@@ -0,0 +1,120 @@
...
+    pullPolicy: IfNotPresent
+    pullSecrets: [ ]
+    repository: syeklu/ckan
+    tag: 0.1.0
+    digest: ""
+  ingress:
+    enabled: true
+    className: ""
+    annotations:
+      # kubernetes.io/ingress.class: nginx
```
cwrau, line 35: This is deprecated anyways

```suggestion
```

charts/ckan/values.yaml
```diff
@@ -0,0 +1,120 @@
...
+      kubernetes.io/tls-acme: "true"
+    baseDomain: ckan.local
```
cwrau, line 37: Stuff like `ingress.host` shouldn't be provided by default

charts/ckan/values.yaml
```diff
@@ -0,0 +1,120 @@
...
+  persistence:
+    accessMode: ReadWriteOnce
+    storageClass: ""
+    size: 100Mi
+    storagePath: "/var/lib/ckan/default"
+  sysadmin:
+    name: ckan_admin
+    password: test1234
```
cwrau, line 45: Stuff like names, passwords!!!, emails, ... shouldn't be set by default as well

charts/ckan/values.yaml
cwrau, line 1: For all permutations of parameters you should create `ci/${testCase?}-values.yaml` or `ci/${testCase?}-gen-values.sh` files, see https://github.com/teutonet/teutonet-helm-charts/tree/main/charts/base-cluster/ci

charts/ckan/templates/ingress.yaml
```diff
@@ -0,0 +1,29 @@
+{{- if .Values.ckan.ingress.enabled -}}
```
cwrau, line 1: You should just check for `.Values.ckan.ingress.baseDomain` or in future `.Values.ckan.ingress.host`, why force the user to set 2 fields if one is enough?

charts/ckan/values.schema.json
```diff
@@ -0,0 +1,259 @@
...
+      "properties": {
+        "imageRegistry": {
+          "type": "string",
+          "description": "The global container image proxy, e.g. [Nexus](https://artifacthub.io/packages/helm/sonatype/nexus-repository-manager), this needs to support various registries",
+          "examples": [
+            "nexus.teuto.net"
+          ]
+        },
+        "imagePullSecrets": {
+          "type": "array"
```
cwrau, line 17: Missing `items` schema

charts/ckan/values.schema.json
```diff
@@ -0,0 +1,259 @@
...
+    },
+    "pullSecrets": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "storageClass": {
+      "type": "string",
+      "description": "The storageClass to use for persistence, e.g. for prometheus, otherwise use the cluster default (teutostack-ssd)"
```
cwrau, line 250:
```suggestion
      "description": "The storageClass to use for persistence, otherwise use the cluster default (e.g. teutostack-ssd)"
```

cwrau (changes requested):

charts/ckan/templates/_helpers.tpl
```diff
@@ -0,0 +1,38 @@
...
+{{/*
+Get service port for ssolr.
+*/}}
+{{- define "ckan.solr.service.port" -}}
+8983
+{{- end -}}
+
+{{- define "ckan.helm.chartSpec" -}}
+  {{- include "common.helm.chartSpec" (dict "context" .context "repo" .repo "chart" .chart "prependReleaseName" false) -}}
+{{- end -}}
```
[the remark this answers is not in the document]
SyeKlu: was superfluous, I have removed it

charts/ckan/README.md.gotmpl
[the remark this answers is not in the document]
SyeKlu: is added

charts/ckan/README.md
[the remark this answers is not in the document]
SyeKlu: is removed

charts/ckan/values.schema.json
```diff
@@ -0,0 +1,259 @@
...
+                  "ReadWriteOncePod"
+                ]
+              },
+              "storageClass": {
+                "$ref": "#/$defs/storageClass"
+              },
+              "size": {
+                "$ref": "#/$defs/quantity"
+              },
+              "storagePath":{
```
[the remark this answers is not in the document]
SyeKlu: it is not configurable in the current status. i have removed it and installed it permanently

charts/ckan/templates/ckan/deployment.yaml
```diff
@@ -0,0 +1,110 @@
...
+    spec:
+      {{- include "common.images.renderPullSecrets" ( dict "images" (list .Values.ckan.image) "context" $) | indent 6 }}
+      automountServiceAccountToken: false
+      volumes:
+        - name: "ckan"
+          persistentVolumeClaim:
+            claimName: {{ $claimName }}
+      containers:
+        - name: {{ printf "%s-%s" .Chart.Name $name }}
+          env:
```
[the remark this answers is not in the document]
SyeKlu: they are now all referenced. however, the passwords etc are required in two secrets, so that automatic generation in the secret would be difficult.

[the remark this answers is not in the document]
cwrau: Why does that make it difficult? Can't you just create two secrets?

cwrau:
Missing `.metadata.namespace` on lots of resources.

charts/ckan/templates/ckan/deployment.yaml
```diff
@@ -0,0 +1,110 @@
+{{- $name := "ckan" -}}
+{{- $claimName := printf "%s-%s" (include "common.names.fullname" $) $name -}}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ printf "%s-%s" (include "common.names.fullname" $) $name | quote }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: {{ $name }}
+spec:
```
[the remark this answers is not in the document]
cwrau: How is this resolved?

charts/ckan/templates/ckan/deployment.yaml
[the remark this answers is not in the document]
SyeKlu: there is a secret with the sql scripts for the database configuration and another one with the secrets for the environment variables

charts/ckan/values.yaml
```diff
@@ -0,0 +1,120 @@
...
+  sysadmin:
+    name: ""
+    password: ""
+    email: ""
+  podSecurityContext: {}
+  securityContext: {}
+  resources: {}
+
+datapusher:
+  replicas: 1
```
cwrau, line 54: Defaults should be production-ready, so this should probably be at least 2?

charts/ckan/values.yaml
```diff
@@ -0,0 +1,120 @@
...
+  coreNames:
+    - ckan
+  auth:
+    enabled: false
+    adminUsername: ""
+    adminPassword: ""
+  cloudEnabled: false
+  cloudBootstrap: false
+  collectionReplicas: 1
+  replicaCount: 1
```
cwrau, line 117: Defaults should be production-ready, so this should probably be at least 2?

charts/ckan/values.yaml
```diff
@@ -0,0 +1,120 @@
...
+  allowPrivilegeEscalation: false
+  privileged: false
+  capabilities:
+    drop:
+      - ALL
+  runAsGroup: 1000
+  runAsUser: 1000
+
+ckan:
+  replicas: 1
```
cwrau, line 28: Defaults should be production-ready, so this should probably be at least 2?

charts/ckan/values.schema.json
```diff
@@ -0,0 +1,286 @@
...
+      "postgresql": {
+        "type": "object",
+        "properties": {
+          "enabled": {
+            "type": "boolean"
+          },
+          "auth": {
+            "type": "object",
+            "required": [
+              "postgresPassword",
```
cwrau, line 169: Shouldn't we generate this one?

charts/ckan/values.schema.json
```diff
@@ -0,0 +1,286 @@
...
+                "$ref": "#/$defs/quantity"
+              }
+            }
+          },
+          "sysadmin": {
+            "type": "object",
+            "additionalProperties": false,
+            "required": [
+              "name",
+              "password",
```
cwrau, line 92: Shouldn't we generate this one?

cwrau (changes requested):
`.metadata.namespace` is missing from some resources

charts/ckan/templates/ckan/deployment.yaml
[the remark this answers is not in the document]
cwrau: Ah, you could create a `.sh` script instead of an `.sql` in which you can use the environment variables with a HEREDOC, see https://stackoverflow.com/a/70976611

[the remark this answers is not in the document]
SyeKlu: is fully implemented for all secrets throughout the chart

charts/ckan/values.yaml
```diff
@@ -0,0 +1,171 @@
...
+        valueFrom:
+          secretKeyRef:
+            name: '{{ printf "%s-config" (include "common.names.fullname" .) }}'
+            key: datastorePassword
+      - name: CKANDB_USER
+        valueFrom:
+          secretKeyRef:
+            name: '{{ printf "%s-config" (include "common.names.fullname" .) }}'
+            key: ckanDatabaseUsername
+      - name: CKANDB_Password
```
cwrau, line 116:
```suggestion
      - name: CKANDB_PASSWORD
```

charts/ckan/templates/ckan/deployment.yaml
```diff
@@ -0,0 +1,166 @@
...
+  replicas: 1
+  strategy:
+    type: RollingUpdate
+  selector:
+    matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
+      app.kubernetes.io/component: {{ $name }}
+  template:
+    metadata:
```
cwrau, line 18: Missing annotation with a secret checksum for auto-rotation on secret changes, see https://github.com/teutonet/teutonet-helm-charts/blob/main/charts/base-cluster/templates/monitoring/kube-prometheus-stack/_grafana-config.yaml#L21

charts/ckan/templates/ckan/secret.yaml
```diff
@@ -0,0 +1,19 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ printf "%s-%s-config" (include "common.names.fullname" $) $name | quote }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: {{ $name }}
+  namespace: {{ .Release.Namespace | quote }}
+stringData:
+  sysAdminUsername: {{ .Values.ckan.sysadmin.name | default "ckan_admin" | quote }}
+  sysAdminPassword: {{ .Values.ckan.sysadmin.password | default (randAlphaNum 42) | quote }}
```
cwrau, line 11: This will result in changing the password *every time*, see https://github.com/teutonet/teutonet-helm-charts/blob/main/charts/base-cluster/templates/monitoring/kube-prometheus-stack/grafana-secret.yaml#L15 on how to do it persistently

charts/ckan/templates/ckan/secret.yaml
```diff
@@ -0,0 +1,19 @@
...
+  sysAdminEmail: {{ .Values.ckan.sysadmin.email | default "admin@test.com" | quote }}
+  {{- if .Values.ckan.smtp }}
+  smtpPassword: {{ .Values.ckan.smtp.password | quote }}
+  {{- end }}
+  {{- if .Values.solr.auth.enabled }}
+  solrUsername: {{ .Values.solr.auth.adminUsername }}
+  solrPassword: {{ .Values.solr.auth.adminPassword }}
```
cwrau, line 18: I would say this should be calculated for the users' convenience

charts/ckan/templates/postgresql/secret.yaml
```diff
@@ -0,0 +1,31 @@
+{{- $name := "postgresql" -}}
+{{- $postgresPassword := (.Values.postgresql.ckanDbs.postgresPassword | default (randAlphaNum 42)) }}
+{{- $ckanDatabasePassword := (.Values.postgresql.ckanDbs.ckanPassword | default (randAlphaNum 42) )}}
+{{- $datastorePassword := (.Values.postgresql.ckanDbs.datastorePassword | default (randAlphaNum 42)) }}
+{{- $datapusherPassword := (.Values.postgresql.ckanDbs.datapusherPassword | default (randAlphaNum 42) )}}
```
cwrau, line 5: These, as well, will be generated anew *every time*

charts/ckan/templates/postgresql/secret.yaml
```diff
@@ -0,0 +1,31 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ printf "%s-config" (include "ckan.postgresql.fullname" .) | quote }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: {{ $name }}
+  namespace: {{ .Release.Namespace | quote }}
+stringData:
+  postgresPassword: {{ $postgresPassword | quote }}
+  replicationPassword: {{ (.Values.postgresql.ckanDbs.replicationPassword | default (randAlphaNum 42)) | quote }}
```
cwrau, line 19: Shouldn't this be handled the same as the ones above?

charts/ckan/templates/postgresql/secret.yaml
```diff
@@ -0,0 +1,31 @@
...
+  ckanDatabaseUsername: {{ $ckanDatabaseUsername | quote }}
+  ckanDatabasePassword: {{ $ckanDatabasePassword | quote }}
+  ckanDatabase: {{ $ckanDatabase | quote }}
+  ckanSqlAlchemyUrl: "postgresql://{{ $ckanDatabaseUsername }}:{{ $ckanDatabasePassword }}@{{ printf "%s-%s" (include "ckan.postgresql.fullname" . ) "primary" }}/{{ $ckanDatabase }}"
```
cwrau, line 23: I would think these URLs are better placed in the corresponding deployment with [dependent variables](https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/#define-an-environment-dependent-variable-for-a-container)

charts/ckan/templates/ckan/deployment.yaml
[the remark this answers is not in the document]
cwrau: If we decide that there can't be 2 pods, then you should set the strategy to `Recreate`

cwrau (changes requested):

charts/ckan/values.yaml
SyeKlu: Unfortunately, this is currently not possible there as there is also the problem with the RWO volume.

charts/ckan/values.yaml
SyeKlu: Unfortunately, this is currently not possible there as there is also the problem with the RWO volume.

charts/ckan/templates/ckan/deployment.yaml
SyeKlu: i have added it there. however, my problem is that i am referencing 3 secrets there. this is not possible with the common.utils.checksumTemplate function

cwrau: That's true, but you can just add multiple annotations 😉

SyeKlu: is now joined as a checksum using `-`

charts/ckan/templates/ingress.yaml
```diff
@@ -0,0 +1,47 @@
...
+  rules:
+    {{- if .Values.ckan.ingress.hostname }}
+    - host: {{ required "You must provide a host, when enabling ingress" .Values.ckan.ingress.hostname | quote }}
+      http:
+        paths:
+          - path: /
+            pathType: Prefix
+            backend: {{- include "common.ingress.backend" (dict "serviceName" $svcName "servicePort" "http" "context" $) | nindent 14 }}
+    {{- end }}
+  {{- if or (and .Values.ckan.ingress.tls (or (include "common.ingress.certManagerRequest" ( dict "annotations" .Values.ckan.ingress.annotations )) .Values.ckan.ingress.selfSigned)) .Values.ckan.ingress.extraTls }}
```
cwrau, line 32:
```suggestion
  {{- if .Values.ckan.ingress.tls }}
```

What is this `extraTls` or this `selfSigned` for?

charts/ckan/templates/ingress.yaml
```diff
@@ -0,0 +1,47 @@
...
+      {{- if .Values.ckan.ingress.existingSecret }}
+      secretName: {{ .Values.ckan.ingress.existingSecret }}
+      {{- else }}
+      secretName: {{ printf "%s-tls" (tpl .Values.ckan.ingress.hostname .) | trunc 63 | trimSuffix "-" }}
+      {{- end }}
+    {{- end }}
+    {{- if .Values.ckan.ingress.extraTls }}
+    {{- toYaml .Values.ckan.ingress.extraTls | nindent 4 }}
+    {{- end }}
+    {{- end }}
```
cwrau, line 46:
```suggestion
    - hosts:
        - {{ tpl .Values.ckan.ingress.hostname . | quote }}
      {{- if .Values.ckan.ingress.existingSecret }}
      secretName: {{ .Values.ckan.ingress.existingSecret }}
      {{- else }}
      secretName: {{ printf "%s-tls" (tpl .Values.ckan.ingress.hostname .) | trunc 63 | trimSuffix "-" }}
      {{- end }}
    {{- end }}
```

cwrau (changes requested):

charts/ckan/templates/ingress.yaml
SyeKlu: i had stuck to the per se standard of bitnami etc. but removed it

cwrau (dismissed):

cwrau (approved):

PR #946, opened 2024-05-13
feat(base-cluster): allow creating git helmRepositories

cwrau:


marvinWolff (approved):

marvinWolff (approved):

tasches:
@cwrau, @marvinWolff is it correct, that the pipeline fails?

cwrau:
The pluto check is ok, we can't do anything about kyverno for now

PR #952, opened 2024-05-13
fix(ci): outputs are always strings 🙄 See https://github.com/actions/runner/issues/1483, again 🤣

cwrau:


marvinWolff:
The linter is failing. 

It says only single quotes should be used.

marvinWolff (approved):

PR #958, opened 2024-05-15
feat(ckan): Add ckan values variable

SyeKlu:
by using the variable for the plugins, the plugins can now be customized by the user. it also allows the use of the DCAT-AP.de plugin from a ckan container.


charts/ckan/values.schema.json
```diff
@@ -63,6 +63,10 @@
           "type": "string",
           "description": "This sets the name of the site, as displayed in the CKAN web interface."
         },
+        "plugins":{
+          "type": "string",
```
cwrau, line 67: This should rather be an array of strings to be more user friendly

charts/ckan/templates/ckan/deployment.yaml
```diff
@@ -138,7 +138,7 @@ spec:
             - name: CKAN__DATAPUSHER__CALLBACK_URL_BASE
               value: "http://{{ printf "%s-%s" (include "common.names.fullname" $) "ckan" }}:{{ include "ckan.ckan.service.port" $ }}/"
             - name: CKAN__PLUGINS
-              value: "envvars image_view text_view recline_view datastore datapusher"
+              value: {{ .Values.ckan.plugins | quote }}
```
cwrau, line 141:
```suggestion
              value: {{ .Values.ckan.plugins | join " " | quote }}
```

charts/ckan/values.yaml
```diff
@@ -26,6 +26,7 @@ defaultSecurityContext: &defaultSecurityContext
   runAsUser: 1000
 
 ckan:
+  plugins: "envvars image_view text_view recline_view datastore datapusher"
```
cwrau, line 29:
```suggestion
  plugins: 
    - envvars
    - image_view
    - text_view
    - recline_view
    - datastore
    - datapusher
```

cwrau (changes requested):

charts/ckan/values.schema.json
SyeKlu: you are right it is the nicer version. i have adapted it

charts/ckan/values.schema.json
```diff
@@ -63,6 +63,16 @@
           "type": "string",
           "description": "This sets the name of the site, as displayed in the CKAN web interface."
         },
+        "plugins":{
+          "type": "array",
+          "items": {
+            "items": {
+              "type": "string"
+            },
+            "additionalProperties": false
```
cwrau, line 72: This is not needed

```suggestion
            }
```

cwrau (changes requested):

charts/ckan/values.schema.json
```diff
@@ -63,6 +63,16 @@
           "type": "string",
           "description": "This sets the name of the site, as displayed in the CKAN web interface."
         },
+        "plugins":{
+          "type": "array",
+          "items": {
+            "items": {
+              "type": "string",
+              "additionalProperties": false
```
cwrau, line 71:
```suggestion
              "type": "string"
```

cwrau (approved):

cwrau (approved):

cwrau (approved):

tasches (approved):

PR #962, opened 2024-05-15
feat(t8s-cluster): add support for HCP

cwrau:


charts/t8s-cluster/templates/workload-cluster/cloud-controller-manager.yaml
```diff
@@ -41,14 +41,33 @@ spec:
...
+  {{- if .Values.controlPlane.hosted }}
+  postRenderers:
+    - kustomize:
+        patches:
+          - target:
+              kind: DaemonSet
+            patch: |
+              - op: replace
+                path: /kind
+                value: Deployment
```
marvinWolff, line 53: Why make the Daemonset a deployment?

cwrau: Because we don't have anything to run the cloud-controller-manager on, especially not a control-plane

Therefore it needs to run on the compute-plane and doesn't need to run on every compute node

marvinWolff: Alrighty

charts/t8s-cluster/templates/management-cluster/clusterClass/openStackMachineTemplates/openStackMachineTemplates.yaml
```diff
@@ -1,9 +1,10 @@
 {{- range $name := list "compute-plane" "control-plane" }}
-{{- if false }}
+  {{- if or (ne $name "control-plane") (not $.Values.controlPlane.hosted) }}
+    {{- if false }}
```
tasches, line 3: I guess you can remove everything between this if statement or do you use still need it for testing purposes?

cwrau: This is not for testing, this is for IntelliJ, as with every other CR we're using with dynamic apiVersion

marvinWolff (approved):

marvinWolff (dismissed):

marvinWolff (approved):

PR #974, opened 2024-05-23
fix(base-cluster): fix nodeCollector is not scheduleable

tasches:


marvinWolff:
The changes seem to be fine, we've tested them. 

Could you check the ci?

marvinWolff (approved):

PR #1024, opened 2024-07-04
fix(base-cluster/monitoring): don't configure route without receiver

cwrau:


charts/base-cluster/templates/monitoring/kube-prometheus-stack/_alertmanager-config.yaml
```diff
@@ -33,36 +33,40 @@ config:
...
+      )
+    -}}
+  {{- end -}}
+  {{- $receiversList := list -}}
+  {{- range $name, $receiver := $receivers -}}
+    {{- $receiversList = append $receiversList (mustMerge (dict "name" $name) $receiver) -}}
+  {{- end }}
+  receivers: {{- $receiversList | toYaml | nindent 4 }}
+  route: {{- if hasKey $receivers "pagerduty" }}
     receiver: pagerduty
```
tasches, line 64: I guess you have to move it in the pagerduty.enabled block

tasches (changes requested):

charts/base-cluster/templates/monitoring/kube-prometheus-stack/_alertmanager-config.yaml
cwrau: It is? It's inside `{{- if hasKey $receivers "pagerduty" }}`

marvinWolff (approved):

tasches (approved):

PR #1052, opened 2024-07-29
fix(base-cluster/kube-prometheus-stack): only register healthcheck if enabled on both sides

cwrau:


tasches (changes requested):
there is a directory included which should not be there.
**crossplane-anynines/**

charts/base-cluster/templates/monitoring/deadMansSwitch/registration.yaml
```diff
@@ -59,8 +63,8 @@ spec:
                 curl --silent --show-error --fail --retry 5 --max-time 30 "https://hc-ping.com/$PING_KEY/$checkName"
```
tasches, line 63: the line above. I tried to setup myself and it did not work. So I made the API call myself and got the Error 400 Bad Request. After some searching in deeper I received the error that something is incorrect with the channel. So I deleted the channel part from the request and it worked. Maybe I did something wrong in the first place of the configuration or this is a bug

tasches (changes requested):

charts/base-cluster/templates/monitoring/deadMansSwitch/registration.yaml
tasches: an integration in healthchecks.io with the global cluster name must be configured. (According to the documentation)

tasches (approved):

marvinWolff (approved):

marvinWolff (approved):

PR #1053, opened 2024-07-29
fix(base-cluster/kube-prometheus-stack): only disable PrometheusNotConnectedToAlertmanagers when Alertmanager is disabled as well

cwrau:


charts/base-cluster/templates/monitoring/kube-prometheus-stack/_alertmanager-config.yaml
```diff
@@ -1,10 +1,15 @@
-{{- define "base-cluster.prometheus-stack.alertmanager.config" -}}
-{{- $enabled := false -}}
-{{- range $_, $receiver := .Values.monitoring.prometheus.alertmanager.receivers -}}
-  {{- $enabled = or $enabled $receiver.enabled -}}
+{{- define "base-cluster.prometheus-stack.alertmanager.enabled" -}}
+  {{- $enabled := false -}}
+  {{- range $_, $receiver := .Values.monitoring.prometheus.alertmanager.receivers -}}
+    {{- $enabled = or $enabled $receiver.enabled -}}
+  {{- end -}}
+  {{- $enabled | ternary true "" }}
```
marvinWolff, line 6: Why `true` or empty and not `true` or `false?`

cwrau: Because helm template works with truthy values:

`{{ if "false" }} this is rendered {{ end }}`
`{{ if "" }} this is not rendered {{ end }}`

marvinWolff: Alright :+1:

marvinWolff (approved):

PR #1062, opened 2024-08-01
fix(base-cluster/kdave): image 2.x.x is unsupported by the helm chart

marvinWolff:
The kdave helm chart does not support image version 2.x.x.

So either we use 1.x.x (probably 1.0.3 as it's the default) or we disable kdave alltogether.


charts/base-cluster/templates/monitoring/kdave/kdave.yaml
```diff
@@ -24,7 +24,7 @@ spec:
     helmBinary: helm3
     image:
       repository: {{ printf "%s/aelbakry/kdave-server" ($.Values.global.imageRegistry | default (include "base-cluster.defaultRegistry" (dict))) }}
-      tag: 2.1.5
+      tag: 1.0.3
```
cwrau, line 27:
```suggestion
      tag: 1.0.4
```

tasches (approved):

PR #1169, opened 2024-09-03
chore(ci): simplify check

cwrau:


marvinWolff:
The check seems to run. Locally it fails because i can't pull the image.

marvinWolff (approved):

PR #1170, opened 2024-09-03
chore(ci): rename job to be more accurate

cwrau:


tasches (dismissed):

tasches (approved):
pipeline errors can be ignored in this case

PR #1175, opened 2024-09-11
fix(ci): run wait-for-checks always

cwrau:


.github/workflows/wait-for-checks.yaml
```diff
@@ -2,6 +2,7 @@ name: Wait for checks
 
 on:
   pull_request: {}
+  pull_request_target: {}
```
marvinWolff, line 5: How does this change make `wait-for-checks` run always?

marvinWolff: This was cleared in a meeting. Its good :+1:

marvinWolff (approved):

PR #1201, opened 2024-09-30
fix(t8s-cluster): create separate etcd defrag jobs per cluster

cwrau:


marvinWolff:
This is ok, but why? :sweat_smile:

cwrau:
> This is ok, but why? 😅

Because otherwise we wouldn't defrag all clusters. In fact, the way it is currently we can't even install a second cluster as this resource is conflicting

tasches (approved):

PR #1203, opened 2024-10-09
feat(ci/license-check): check if specifed images have been manually l…

marvinWolff:
…icense checked

Also fixed title of job.

This will fix the problem with trivy, that the licenses of the software in an docker-image are ok, and not if the docker-image itself is ok.

The list will contain every image used by us and check, if an image used by us is not in this list.


.github/workflows/check-licenses.yaml
```diff
@@ -23,3 +23,18 @@ jobs:
...
       - run: ./.github/scripts/scan-for-licenses.sh ${{ needs.getChangedChart.outputs.chart }}
+  check-licenses-list:
+    name: check licenses from list
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+      - run: pip install yq
+      - run: |
+          mapfile -t IMAGES < <(for chart in charts/*; do if [[ -f "$chart/Chart.yaml" ]]; then  cat $chart/Chart.yaml | yq -r '.annotations["artifacthub.io/images"] // ""'; fi; done | cut -d ":" -f2 | uniq | sort | sed '/^$/d')
+          # shellcheck disable=SC2068
```
marvinWolff, line 34: When i quote the array the whole array will be expanded, i can't figure this out. Maybe you have an idea?

.github/workflows/check-licenses.yaml
```diff
@@ -23,3 +23,18 @@ jobs:
...
       - run: pip install yq
       - run: ./.github/scripts/scan-for-licenses.sh ${{ needs.getChangedChart.outputs.chart }}
+  check-licenses-list:
+    name: check licenses from list
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+      - run: pip install yq
+      - run: |
+          mapfile -t IMAGES < <(for chart in charts/*; do if [[ -f "$chart/Chart.yaml" ]]; then yq -r '.annotations["artifacthub.io/images"] // ""' <"$chart/Chart.yaml" ; fi; done | cut -d ":" -f2 | uniq | sort | sed '/^$/d')
```
cwrau, line 33: This should rather just use the $chart from `getChangedChart`

.github/scripts/licenseList
cwrau, line 1: This should rather be in `.github` and be a proper file like yaml

cwrau (changes requested):

.github/workflows/check-licenses.yaml
marvinWolff: I hope this is fine now, testing is hard when using github-ci

.github/workflows/check-licenses.yaml
```diff
@@ -23,3 +23,19 @@ jobs:
...
+    name: check licenses from list
+    runs-on: ubuntu-latest
+    needs: getChangedChart
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+      - run: pip install yq
+      - run: |
+          mapfile -t IMAGES < <(if [[ -f "${{ needs.getChangedChart.outputs.chart }}/Chart.yaml" ]]; then yq -r '.annotations["artifacthub.io/images"] // ""' <"${{ needs.getChangedChart.outputs.chart }}/Chart.yaml" ; fi | cut -d ":" -f2 | uniq | sort | sed '/^$/d')
+          #shellcheck disable=SC2068
+          mapfile -t RESULT < <(for IMAGE in ${IMAGES[@]}; do grep -q "$IMAGE" ./.github/licenseList.yaml || (echo "$IMAGE";); done;)
```
cwrau, line 36: Please really parse the yaml, don't grep it

.github/licenseList.yaml
```diff
@@ -0,0 +1,211 @@
+licenses:
```
cwrau, line 1: This should rather be a map;

```yaml
licenses:
  $image:
    license: $license
    licenseLink: $link
```

This would also prevent accidentally adding an image duplicate

cwrau (changes requested):

.github/licenseList.yaml
marvinWolff: This doesn't work.

Map keys must be unique

tasches (dismissed):

marvinWolff:
This is broken. See:  
https://github.com/teutonet/teutonet-helm-charts/actions/runs/11912138960/job/33195123294?pr=1239

and 

https://github.com/teutonet/teutonet-helm-charts/actions/runs/11912138960/job/33195123698?pr=1239

PR #1249, opened 2024-11-21
feat(ckan): add api creation and perist secrets (#1238)

marvinWolff:


tasches (approved):

SyeKlu (approved):

charts/ckan/templates/ckan/configmap.yaml
```diff
@@ -0,0 +1,76 @@
+{{- $name := (printf "%s-%s-configmap" (include "common.names.fullname" .) "ckan") -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ $name }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: {{ $name }}
```
cwrau, line 7: Is this even a component and not just ckan?

Also, don't forget to add the namespace

```suggestion
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ include "common.names.fullname" . }}
  namespace: {{ .Release.Namespace }}
  labels: {{- include "common.labels.standard" $ | nindent 4 }}
```

cwrau (approved):

charts/ckan/templates/ckan/configmap.yaml
SyeKlu: However, the name no longer has the configmap ending, does it? And is it then more difficult to distinguish between secret and configmap in the templates?

SyeKlu: i will optimise the naming concept in a followup

PR #1293, opened 2025-01-21
chore(base-cluster): make proxy-protocol configurable

SyeKlu:
in order to operate baremetal clusters with bgp loadbalancing without an extra upstream proxy, this option would have to be deactivatable, otherwise the ngnix would not be able to process the incoming traffic


charts/base-cluster/values.schema.json
```diff
@@ -1308,6 +1308,19 @@
...
+            "config":{
+              "type": "object",
+              "properties": {
+                "enabled": {
+                  "useProxyProtocol": "boolean"
+                }
+              }
+            }
+          }
+        },
```
cwrau, line 1323:
```suggestion
        "useProxyProtocol": {
          "type": "boolean"
        },
```

charts/base-cluster/values.yaml
```diff
@@ -425,6 +425,9 @@ ingress:
   resources:
     limits:
       cpu: 1
+  controller:
+    config:
+      useProxyProtocol: true
```
cwrau, line 430:
```suggestion
  useProxyProtocol: true
```

charts/base-cluster/templates/ingress/nginx.yaml
```diff
@@ -36,7 +36,7 @@ spec:
         {{- end }}
       {{- end }}
       config:
-        use-proxy-protocol: true
+        use-proxy-protocol: {{ .Values.ingress.controller.config.useProxyProtocol }}
```
cwrau, line 39:
```suggestion
        use-proxy-protocol: {{ .Values.ingress.useProxyProtocol }}
```

cwrau (changes requested):

cwrau (approved):

cwrau:
@SyeKlu You missed https://github.com/teutonet/teutonet-helm-charts/pull/1293/files#diff-9cc4f2f70b580ff862ae4b206f7feabf9ba7c06563f0ff0c1e48e1b274d8293eR50, does it still work? 🤔

PR #1310, opened 2025-01-22
feat(teuto-cnpg): add cnpg-wrapper skel

marvinWolff:


charts/teuto-cnpg/Chart.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v2
+name: teuto-cnpg
+description: A Helm chart for Kubernetes
```
cwrau, line 3: A better description would be nice 😁

charts/teuto-cnpg/templates/podmonitor.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    cnpg.io/cluster: {{ printf "%s" .Release.Name }}
+    managed.by/monitoring: teutonet
```
cwrau, line 6: This should be configurable? like it is in https://github.com/teutonet/teutonet-helm-charts/blob/main/charts/base-cluster/values.yaml#L219-L222

That way the user is not forced to use our base-cluster / this specific label

charts/teuto-cnpg/values.schema.json
```diff
@@ -0,0 +1,108 @@
...
+            "15.6",
+            "16.1",
+            "17.2"
+          ]
+        },
+        "digest": {
+          "type": "string"
+        }
+      },
+      "additionalProperties": false
```
cwrau, line 105: Shouldn't `roles` and `databases` be required fields?

charts/teuto-cnpg/ci/artifacthub-values.yaml
cwrau, line 1: And after https://github.com/teutonet/teutonet-helm-charts/pull/1310/files#r1932104527 this would need `databases` set

charts/teuto-cnpg/values.schema.json
marvinWolff: I'm not sure. I mean, you probably want them if this chart is used, but you don't need to i guess?

marvinWolff:
Done the suggestion.

cwrau (approved):

PR #1347, opened 2025-02-20
feat(base-cluster/monitoring)!: migrate promtail to alloy

cwrau:
- update loki to an up-to-date version
- also replace otel collector with alloy, saving resources!


marvinWolff:
CI fails.

tasches (changes requested):

tasches:
The scaling option for loki is dropped so I guess this is a breaking change right?

cwrau:
> The scaling option for loki is dropped so I guess this is a breaking change right?

That's not really a breaking change. But this PR is a breaking change in any case.

tasches (approved):

PR #1414, opened 2025-03-11
feat(base-cluster): use new networkPolicy template

cwrau:


marvinWolff (approved):

tasches (dismissed):
I tried it locally as well. The pipeline failed with `error calling include: template: no template "common.networkPolicy.type"` and I can not find a function with this name in our common chart.

marvinWolff (approved):

PR #1417, opened 2025-03-17
feat(base-cluster/ingress-nginx): use risk-level Critical when annotations are enabled

ol-teuto:
This sets `annotations-risk-level` to Critical when nginx configuration snippets are enabled, restoring the default setting from before ingress-nginx release 1.12.

## Summary by CodeRabbit

- **New Features**
	- Introduced a conditional risk level configuration that automatically applies a "Critical" status when snippet-based annotations are enabled, enhancing overall risk management without affecting other existing settings.

### DIFF
--- a/charts/base-cluster/templates/ingress/nginx.yaml
+++ b/charts/base-cluster/templates/ingress/nginx.yaml
@@ -40,6 +40,9 @@ spec:
         use-proxy-protocol: {{ .Values.ingress.useProxyProtocol }}
         use-gzip: true
         enable-brotli: true
+        {{- if .Values.ingress.allowNginxConfigurationSnippets }}
+        annotations-risk-level: Critical
+        {{- end }}
         enable-underscores-in-headers: true
         {{- if $telemetryConf.enabled }}
         enable-opentelemetry: true


charts/base-cluster/templates/ingress/nginx.yaml
```diff
@@ -40,6 +40,9 @@ spec:
         use-proxy-protocol: {{ .Values.ingress.useProxyProtocol }}
         use-gzip: true
         enable-brotli: true
+        {{- with .Values.ingress.annotationsRiskLevel }}
```
cwrau, line 43: As we're trying to move away from provider-specific configuration, I'd rather "merge" this with the existing [config to allow for configuration snippets](https://github.com/teutonet/teutonet-helm-charts/pull/1417/files#diff-9cc4f2f70b580ff862ae4b206f7feabf9ba7c06563f0ff0c1e48e1b274d8293eR22) instead of providing more specific configuration.

Meaning, if that is enabled use `critical`, otherwise "nothing" to use the nginx charts default.

If possible [the user shouldn't use any of this anyways](https://github.com/teutonet/teutonet-helm-charts/pull/1417/files#diff-f8f80a4e01e956f12eaf3fb3377ba752c07494cb32c795908c68ce91df075c0dR1336), but legacy is legacy...

ol-teuto: I wanted to avoid a breaking change, but I think it would make sense to automatically enable snippets when this is set

cwrau: Why would this be a breaking change? 🤔

I'd say it's quite the opposite, with this we're fixing the breaking change in a nginx minor release (them breaking semver) and keeping our users on a working configuration without them needing to do something.

Doing it your proposed way would be a breaking change, as now the user would need to configure another field.

Why not the other way around? There are only five `critical` annotations, of which we only support one, the `configurationSnippets`

So, if the user wants to use the configuration snippets they have to set two fields, both the enable field and the annotation risk level field.

If so, then we'd also need a validation to check that if the snippets are enable the risk level is also correctly set, as the other way around is an invalid configuration, at least semantically.

cwrau (changes requested):

charts/base-cluster/templates/ingress/nginx.yaml
ol-teuto: I just misunderstood what you said, using critical by default is fine by me

ol-teuto: I just thought it makes sense to give users more options, in case they want to enable some annotations, but not the high risk ones. But if there is no need for that, we don't need to implement it

PR #1420, opened 2025-03-19
feat(base-cluster/ingress)!: add option traefik for ingress controller

cwrau:


charts/base-cluster/templates/ingress/traefik.yaml
```diff
@@ -0,0 +1,89 @@
...
+  interval: 1h
+  driftDetection:
+    mode: enabled
+  {{- if .Values.monitoring.prometheus.enabled }}
+  dependsOn:
+    - name: kube-prometheus-stack
+      namespace: monitoring
+  {{- end }}
+  install:
+    crds: Skip
```
marvinWolff, line 21: Where do the CRDs get installed?

I know there is an extra Chart, but i can't find it here

cwrau: CRDs are explicitly not installed to encourage users to use standard APIs like Ingress and Gateway API instead of custom resources/configurations/annotations/... .

marvinWolff (approved):

PR #1429, opened 2025-03-26
feat(base-cluster/monitoring): remove unnecessary labels from monitoring components

cwrau:
that way the user doesn't need to add labels to their \*Monitors

for internal alert routing we should select on `monitoring/managed-by: teutonet`

## Summary by CodeRabbit

- **Refactor**
  - Standardized monitoring configurations by replacing dynamic label assignments with a fixed label ("monitoring/provisioned-by: base-cluster") across various monitoring resources to ensure consistent identification.

- **Chore**
  - Simplified settings by removing obsolete label selectors and redundant monitoring configuration properties from schemas and values, streamlining resource management and future customization.

### DIFF
--- a/charts/base-cluster/templates/backup/velero.yaml
+++ b/charts/base-cluster/templates/backup/velero.yaml
@@ -90,11 +90,11 @@ spec:
       uploaderType: restic
     metrics:
       serviceMonitor:
-        additionalLabels: {{- include "common.tplvalues.render" (dict "value" .Values.monitoring.labels "context" .) | nindent 10 }}
+        additionalLabels:
+          monitoring/managed-by: teutonet
         enabled: true
       prometheusRule:
         enabled: true
-        additionalLabels: {{- include "common.tplvalues.render" (dict "value" .Values.monitoring.labels "context" .) | nindent 10 }}
         spec:
           - alert: VeleroBackupFailures
             annotations:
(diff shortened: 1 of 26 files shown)


marvinWolff:
This would be a breaking change.

We don't know who uses this for what reasons. Simply removing this might break things.

Why even remove the flag altogether? We can just add the label hardcoded still keep the option to set labels.

cwrau:
> This would be a breaking change.
> 
> We don't know who uses this for what reasons. Simply removing this might break things.

If you really want to make this a breaking change we can do that.

I've checked for all our customers and no one is using that. As we also only remove this field, all the monitoring still works, at worst there are just more metrics / alerts.

> Why even remove the flag altogether? We can just add the label hardcoded still keep the option to set labels.

Because we A) don't need it and B) don't want it.

We have misunderstood the usage of this chart in this aspect.
If you run this chart for yourself this field is incredibly annoying and useless.
If you run this chart for someone else, like we do, you still don't want this, as the customer wants to reuse this monitoring infrastructure and you need to filter on the sending side anyways.

marvinWolff:
lgtm but the nitpick here should probably be fixed:

```
YAMLLint reports an indentation mismatch here. Ensure that each label key under labels: is indented exactly two spaces to conform with YAML best practices.
```

marvinWolff (dismissed):

cwrau:
@marvinWolff sorry for dismissing the review again, I forgot that there was still `teutonet` written here and traefik had to be adjusted as well

marvinWolff (approved):

PR #1458, opened 2025-05-13
chore(base-cluster/velero): add another alert that won't fail when th…

marvinWolff:
…e pod restarts

### DIFF
--- a/charts/base-cluster/templates/backup/velero.yaml
+++ b/charts/base-cluster/templates/backup/velero.yaml
@@ -103,4 +103,11 @@ spec:
             for: 15m
             labels:
               severity: warning
+          - alert: VeleroBackupOlderThan30Hours
+            annotations:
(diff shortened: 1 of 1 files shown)


tasches (approved):

charts/base-cluster/templates/backup/velero.yaml
```diff
@@ -103,4 +103,11 @@ spec:
             for: 15m
             labels:
               severity: warning
+          - alert: VeleroBackupOlderThan30Hours
```
cwrau, line 106: What about backups that should only be run every week? Or backups that should run every 6 hours?

I don't think it makes sense to have a global alert for this, as every backup is different.

While looking up the specifics of this, I couldn't find this metric? Who exposes it?

cwrau (changes requested):

PR #1480, opened 2025-05-22
feat(t8s-cluster): add annotation in preparation for etcd-backup with…

marvinWolff:
… k8sup

## Summary by CodeRabbit

- **New Features**
  - Added an annotation patch for the etcd component to include a backup command using etcdctl with TLS certificates.


marvinWolff:
After much research and back and forth we can't use it this way.

We can't output the backup to stdout because of missing tools to do that. We will now use pre-backup pods for that.

PR #1494, opened 2025-05-30
feat(base-cluster/trivy): add the option to add labels to all trivy reports

tasches:
This is for example needed if you want to exclude trivy reports from a velero backup.

see `https://github.com/aquasecurity/trivy-operator/blob/main/deploy/helm/README.md?plain=1#L195C3-L195C16`

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying additional report labels for Trivy operator reports through an optional configuration parameter. This allows users to customize report labeling in security scanning settings.


cwrau:
Why exactly is this needed? Couldn't you just not back all of these reports up instead of excluding these by label?

tasches:
It is possible to exclude resources in the schedule. See velero doc
https://velero.io/docs/v1.3.0/api-types/schedule/. Therefore this pr is no longer needed

PR #1554, opened 2025-07-03
fix(ci/scan-for-cves): continue on failure, else the issue job never …

marvinWolff:
…runs

## Summary by CodeRabbit

* **Chores**
  * Updated workflow to allow the process to continue even if errors occur during security scanning.


marvinWolff:
This will be fixed by the next PR from Sven.

PR #1640, opened 2025-07-30
feat(urbanstack-shared-app): add urbanstack-shared-app chart

ol-teuto:
Will be used by urbanstack to deploy shared apps. Most values will be set by the platform, only the image can be chosen by the user.


charts/urbanstack-shared-app/ci/no-auth-values.yaml
cwrau, line 1: An empty ci-values file is not needed, that's the same as the normal values.yaml

charts/urbanstack-shared-app/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,59 @@
...
+          port: 443
+  ingress:
+    - from: # allow nginx proxy to call this
+        - podSelector:
+            matchLabels:
+              app.kubernetes.io/instance: ingress-nginx
+              app.kubernetes.io/name: ingress-nginx
+          namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: ingress-nginx
```
cwrau, line 59: This is hardcoded to ingress-nginx, what if the user doesn't have another ingress controller?

charts/urbanstack-shared-app/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,59 @@
...
+      ports:
+        - protocol: TCP
+          port: 80
+        - protocol: TCP
+          port: 443
```
cwrau, line 50: Are you sure this works?

So the application (hardcoded) calls the ingress-nginx but with a different Host header / SNI so (hardcoded) nginx proxies it correctly?

charts/urbanstack-shared-app/templates/ingress.yaml
```diff
@@ -0,0 +1,26 @@
...
+            pathType: Prefix
+            backend:
+              service:
+                name: {{ .Release.Name | quote }}
+                port:
+                  name: http
+  tls:
+  - hosts:
+    - {{ .Values.host | quote }}
+    secretName: {{ printf "%s-tls" .Values.host | quote }}
```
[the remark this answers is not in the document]
ol-teuto: The installation will fail so not necessary imo

charts/urbanstack-shared-app/templates/_helpers.tpl
```diff
@@ -0,0 +1,32 @@
...
+{{/*
+Common labels
+*/}}
+{{- define "urbanstack-shared-app.labels" -}}
+helm.sh/chart: {{ include "urbanstack-shared-app.chart" . | quote }}
+tenant: {{ .Values.tenant | quote }}
+citytool: {{ .Values.citytool | quote }}
+{{ include "urbanstack-shared-app.selectorLabels" . }}
+app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+{{- end }}
```
cwrau, line 24: You should use https://artifacthub.io/packages/helm/bitnami/common#labels via our [common chart](https://github.com/teutonet/teutonet-helm-charts/tree/main/charts/common)

charts/urbanstack-shared-app/templates/_helpers.tpl
```diff
@@ -0,0 +1,32 @@
...
+
+{{/*
+Selector labels
+*/}}
+{{- define "urbanstack-shared-app.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "urbanstack-shared-app.name" . | quote }}
+app.kubernetes.io/instance: {{ .Release.Name | quote }}
+{{- end }}
```
cwrau, line 32: Same here, use https://artifacthub.io/packages/helm/bitnami/common#labels

cwrau (changes requested):

charts/urbanstack-shared-app/values.yaml
```diff
@@ -0,0 +1,35 @@
+image:
+  repository: ""
+  registry: ""
+  sha: ""
```
cwrau, line 4: Missing image?

charts/urbanstack-shared-app/templates/_helpers.tpl
ol-teuto: Is it really worth it to pull in a dependency for this?

charts/urbanstack-shared-app/values.yaml
ol-teuto: no, the image is supplied by the user

charts/urbanstack-shared-app/templates/networkpolicy.yaml
ol-teuto: I think it should work, nginx has to handle the Host header in normal scenarios too. This is also only needed for local testing, so I could add an option to enable/disable this

charts/urbanstack-shared-app/templates/networkpolicy.yaml
ol-teuto: And if we ever switch ingress we just need to update this chart

PR #1643, opened 2025-07-31
feat(ci): install gh for self hosted runners

tasches:


.github/workflows/wait-for-checks.yaml
```diff
@@ -17,6 +17,8 @@ jobs:
     permissions:
       checks: read
     steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
```
cwrau, line 20: Why are you checking something out?

tasches: It was a try to get the brew bin but unfortunately it did not have the desired effect

PR #2020, opened 2026-03-13
feat(t8s-cluster/rbac): use Users instead of the whole Group

cwrau:
## Summary by CodeRabbit

* **Chores**
  * Updated cluster authorization and role-based access control configuration.


marvinWolff:
This excludes everyone except us from the on-call service. This can't be merged withouth furhter discussion

PR #2089, opened 2026-04-20
feat(ci): only output critical vulnerabilities

marvinWolff:
## Summary by CodeRabbit

* **Bug Fixes**
  * Security scan reports now filter results to show only critical-severity findings and corresponding error-level entries.
  * Report entries are enriched with automation identifiers and chart/sub-chart metadata for clearer reporting and visualization.

### DIFF
--- a/.github/scripts/generate-sarif-reports.sh
+++ b/.github/scripts/generate-sarif-reports.sh
@@ -49,7 +49,7 @@ function generateSarifReport() {
   chartsJson="$(yq --arg image "$image" -r '.annotations["artifacthub.io/images"] | split("\n")[] | select(contains($image))' "$chart/Chart.yaml" |
     awk '{print $NF}' |
     jq -r -c -Rn '[inputs]')"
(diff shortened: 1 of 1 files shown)


.github/scripts/generate-sarif-reports.sh
```diff
@@ -49,7 +49,7 @@ function generateSarifReport() {
   chartsJson="$(yq --arg image "$image" -r '.annotations["artifacthub.io/images"] | split("\n")[] | select(contains($image))' "$chart/Chart.yaml" |
     awk '{print $NF}' |
     jq -r -c -Rn '[inputs]')"
-  if GRYPE_DB_AUTO_UPDATE=false grype "$image" -o sarif --by-cve --only-fixed | jq -r --arg category "$chart/${GITHUB_JOB:-local}" --argjson subCharts "$chartsJson" --arg chart "$chartName" '.runs |= (map(.automationDetails = {id: $category}) | map(.properties = {chart: $chart, subCharts: $subCharts}))' >"$tmpFile"; then
```
cwrau, line 52: Why not just output critical by grype instead of filtering later?

cwrau (changes requested):

.github/scripts/generate-sarif-reports.sh
marvinWolff: Because grype currently does not support that.

Here is an issue where that feature is discussed, but is not currently implemented: https://github.com/anchore/grype/issues/2006

Also there is the option `--fail-on <severity>`, but that will also print out every cve and only will set a return code when the threshold is exceeded.

marvinWolff:
@coderabbitai review

cwrau (approved):

PR #2191, opened 2026-06-12
feat(base-cluster): add extra ports for traefik and gatewayapi

SyeKlu:
## Summary by CodeRabbit

* **New Features**
  * Traefik ingress now supports configurable additional ports for flexible service exposure
  * Select TCP/UDP protocol for each custom port mapping
  * Optional proxy protocol configuration for enhanced port management

### DIFF
--- a/charts/base-cluster/templates/ingress/traefik.yaml
+++ b/charts/base-cluster/templates/ingress/traefik.yaml
@@ -42,6 +42,18 @@ spec:
       websecure:
         proxyProtocol:
           insecure: {{ include "base-cluster.ingress.useProxyProtocol" (dict "context" $) }}
+      {{- range $name, $cfg := .Values.ingress.extraPorts }}
+      {{ $name }}:
+        expose:
+          default: {{ $cfg.expose.default }}
+        exposedPort: {{ $cfg.exposedPort }}
+        port: {{ $cfg.port }}
+        protocol: {{ $cfg.protocol }}
+        {{- with $cfg.proxyProtocol }}
+        proxyProtocol:
+          insecure: {{ .insecure }}
+        {{- end }}
+      {{- end }}
     service:
       annotations:
         loadbalancer.openstack.org/timeout-client-data: "2073600000"
--- a/charts/base-cluster/values.schema.json
+++ b/charts/base-cluster/values.schema.json
@@ -1379,7 +1379,35 @@
           "type": "string",
           "pattern": "^((25[0-5]|(2[0-4]|1\\d|[1-9]|)\\d)\\.?\\b){4}$",
           "description": "Try to use specified IP as loadbalancer IP"
-        }
+        },
+        "extraPorts": {
+              "type": "object",
+              "additionalProperties": {
+                "type": "object",
+                "required": ["port", "exposedPort", "protocol"],
+                "properties": {
+                  "port": { "type": "integer", "minimum": 1, "maximum": 65535  },
+                  "exposedPort": { "type": "integer", "minimum": 1, "maximum": 65535  },
+                  "protocol": {
+                    "type": "string",
+                    "enum": ["TCP", "UDP"]
+                  },
+                  "expose": {
+                    "type": "object",
+                    "properties": {
+                      "default": { "type": "boolean" }
+                    }
+                  },
+                  "proxyProtocol": {
+                    "type": "object",
+                    "properties": {
+                      "insecure": { "type": "boolean" }
+                    }
+                  }
+                },
+                "additionalProperties": true
+              }
+            }
       },
       "additionalProperties": false
     },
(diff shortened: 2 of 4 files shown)


marvinWolff (approved):

charts/base-cluster/values.schema.json
```diff
@@ -1379,7 +1379,35 @@
...
+                  "protocol": {
+                    "type": "string",
+                    "enum": ["TCP", "UDP"]
+                  },
+                  "expose": {
+                    "type": "object",
+                    "properties": {
+                      "default": { "type": "boolean" }
+                    }
+                  },
```
cwrau, line 1400: This is useless, if not exposed, why do this.

charts/base-cluster/values.schema.json
```diff
@@ -1379,7 +1379,35 @@
...
+                  "proxyProtocol": {
+                    "type": "object",
+                    "properties": {
+                      "insecure": { "type": "boolean" }
+                    }
+                  }
+                },
```
cwrau, line 1407: This should just be the same as the other builtin ports, as they use the same service anyways, this can't be mixed.

charts/base-cluster/templates/ingress/traefik.yaml
```diff
@@ -42,6 +42,18 @@ spec:
       websecure:
         proxyProtocol:
           insecure: {{ include "base-cluster.ingress.useProxyProtocol" (dict "context" $) }}
+      {{- range $name, $cfg := .Values.ingress.extraPorts }}
```
cwrau, line 45: It should be validated, that there is no port clash

cwrau, line 45: It should be verified, that the builtin ports won't be overwritten

charts/base-cluster/templates/ingress/traefik.yaml
```diff
@@ -42,6 +42,18 @@ spec:
...
+      {{ $name }}:
+        expose:
+          default: {{ $cfg.expose.default }}
+        exposedPort: {{ $cfg.exposedPort }}
+        port: {{ $cfg.port }}
+        protocol: {{ $cfg.protocol }}
```
cwrau, line 51: TCP should be default

charts/base-cluster/templates/ingress/traefik.yaml
```diff
@@ -42,6 +42,18 @@ spec:
       websecure:
         proxyProtocol:
           insecure: {{ include "base-cluster.ingress.useProxyProtocol" (dict "context" $) }}
+      {{- range $name, $cfg := .Values.ingress.extraPorts }}
+      {{ $name }}:
+        expose:
+          default: {{ $cfg.expose.default }}
+        exposedPort: {{ $cfg.exposedPort }}
```
cwrau, line 49: This should use the `port` as default

charts/base-cluster/values.schema.json
```diff
@@ -1379,7 +1379,35 @@
           "type": "string",
           "pattern": "^((25[0-5]|(2[0-4]|1\\d|[1-9]|)\\d)\\.?\\b){4}$",
           "description": "Try to use specified IP as loadbalancer IP"
-        }
+        },
+        "extraPorts": {
+              "type": "object",
```
cwrau, line 1384: This should also be posible to just be a mapping;

```yaml
extraPorts:
  mqtt: 1883
```

charts/base-cluster/values.schema.json
cwrau, line 1: This should be formatted in a way hat doesn't collapse lines, (no `"required": ["port", "exposedPort", "protocol"],`)

cwrau:
@SyeKlu the review comments should be addressed in a new PR of course.

The new PR will be a breaking change.

charts/base-cluster/values.schema.json
SyeKlu: For example, if you use port 1883, the ‘enable’ setting here blocks the traffic; based on my tests, it is only required for TLS.

cwrau: Are you sure? According to the docs this just allows proxy protocol from all IPs; https://doc.traefik.io/traefik/reference/install-configuration/entrypoints/#opt-proxyProtocol-insecure

And as far as I know, proxy protocol is either on of off for the whole (openstack) loadbalancer

charts/base-cluster/templates/ingress/traefik.yaml
cwrau: These are the pre-existing ones;

```yaml
- containerPort: 9100 
  name: metrics       
  protocol: TCP       
- containerPort: 8080 
  name: traefik       
  protocol: TCP       
- containerPort: 8000 
  name: web           
  protocol: TCP       
- containerPort: 8443 
  name: websecure     
  protocol: TCP       
```

charts/base-cluster/values.schema.json
cwrau: I just tried this with port 1883 and `insecure: true` with an tcp echo server; it worked.

Also tried with mosquitto, also worked.

Both just with a TCPRoute.

PR #2215, opened 2026-07-07
feat(base-cluster/monitoring): add blackbox-exporter

cwrau:
this is required for the `Probes` CRD

## Summary by CodeRabbit

* **New Features**
  * Added optional Prometheus Blackbox Exporter support for monitoring deployments.
  * Added configuration for enabling the exporter and assigning resource settings.
  * Enabled self-monitoring and integrated the exporter with the existing Prometheus stack.
* **Configuration**
  * Added the exporter’s chart configuration and license information.
  * The exporter remains disabled by default and can be enabled through monitoring settings.

### DIFF
--- a/charts/base-cluster/templates/monitoring/kube-prometheus-stack/blackbox-exporter.yaml
+++ b/charts/base-cluster/templates/monitoring/kube-prometheus-stack/blackbox-exporter.yaml
@@ -1,0 +1,29 @@
+{{- if and .Values.monitoring.prometheus.enabled .Values.monitoring.prometheus.blackboxExporter.enabled }}
+apiVersion: helm.toolkit.fluxcd.io/v2
+kind: HelmRelease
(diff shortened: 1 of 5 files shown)


charts/base-cluster/templates/monitoring/kube-prometheus-stack/blackbox-exporter.yaml
```diff
@@ -0,0 +1,29 @@
+{{- if .Values.monitoring.prometheus.enabled }}
```
marvinWolff: I think we should make this more explicit. Most people won't need the blackbox exporter, but this would roll it out on almost every cluster.


Maybe just a value that needs to be set explicitly like `.Values.monitoring.blackbox.enabled`.

What do you think?

marvinWolff (approved):

PR #2392, opened 2026-09-15
fix(base-cluster): bump grafana memory limit to 512Mi

cwrau:
## Summary
- Grafana's `nano` resources preset caps memory at 192Mi, which is too tight — observed repeated OOMKilled/CrashLoopBackOff across many clusters (mgmt clusters and tenant clusters alike), with RSS sitting near/above 192Mi right after a fresh restart.
- Sets an explicit `monitoring.grafana.resources.limits.memory: 512Mi` override (same pattern already used for `monitoring.prometheus.kubeStateMetrics`/`monitoring.loki` in consumer repos), so every cluster picks it up without needing a per-repo override.

## Test plan
- [x] `helm lint charts/base-cluster` passes (pre-existing warnings only)
- [x] `helm template charts/base-cluster` renders `monitoring.grafana.resources.limits.memory: 512Mi` into the kube-prometheus-stack HelmRelease values
- [x] Live-patched `mgmt-bfe-prod`'s grafana Deployment to 512Mi as an immediate mitigation ahead of this rollout


charts/base-cluster/values.yaml
```diff
@@ -276,7 +276,9 @@ monitoring:
     additionalPlugins: []
     notifiers: []
     resourcesPreset: nano
-    resources: {}
+    resources:
+      limits:
+        memory: 512Mi
```
marvinWolff, line 281: This is probably not enough. I would raise this to 1gb

cwrau: Bumped to 1Gi, and made grafana opt-in (`monitoring.grafana.enabled: false` by default) since it now needs a heavier resource commitment — marked as a breaking change.

CLAUDE.md
```diff
@@ -41,6 +41,10 @@ Try to keep this CLAUDE file up to date with any gotchas you encounter while wor
...
 - Prefer one field with a `type`/mode discriminator over multiple independent `enabled` toggles when the options are mutually exclusive ways to do the same thing (e.g. one ingress mechanism, not "ingress OR gatewayApi, both togglable at once").
 
+## Chart lint/test gotchas
+
+- `chart:lint`, `chart:template` and `chart:test` all depend on `chart:prepare-values`, which runs `.github/scripts/prepare-values.sh`. That script merges each chart's `ci/_common.sh` overrides straight into `charts/<chart>/values.yaml` via `sponge` — an in-place, destructive overwrite of the real file (e.g. it stomps `global.clusterName`/`monitoring.grafana.adminPassword` with CI test values), not a copy. It's meant for a disposable CI checkout. Running these tasks against a local working tree permanently mutates `values.yaml`; `git checkout -- charts/<chart>/values.yaml` afterwards to restore it. Prefer plain `helm lint`/`helm template` locally when just sanity-checking a values change.
```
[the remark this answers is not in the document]
cwrau: Fixed both: dropped chart:test from the prepare-values dependency list (it only depends on internal:dependency-update, confirmed in taskfile.yaml), and softened the git-checkout advice to check the diff / restore only the CI-override fields instead of a blind revert.

